PR #2, opened 2016-06-28
Feature/timeline

apercas:
Pretty basic timeline. WIP, but mergeable.
timeline


apercas:
Updated with clickable timeline.

davidsingal:
It looks very good, the code is very good too. Please avoid to use var with ES6.

apercas:
@davidsingal Updated :)

tiagojsag:
There seems to be a difference in scope between `var` and `let`, please test stuff ;)

apercas:
Sure, everything is working as smooth as it did before.

PR #5, opened 2016-06-29
Feature/simple tooltip with dates rebased

tiagojsag:


tiagojsag:
@apercas @davidsingal can you take another look at this, just ot be sure all is here and nothing is lost/messed up? thanks

PR #7, opened 2016-06-29
Add basic auth to production env

tiagojsag:
Add basic auth to production env


config/environments/production.js
```diff
@@ -2,10 +2,12 @@
 
 const path = require('path');
 const express = require('express');
+const basicAuth = require('basic-auth-connect');
 
 const indexPath = path.join(process.cwd(), 'dist/index.html');
 
 module.exports = function(app) {
+  app.use(basicAuth('gfw', 'fishtrack'));
```
davidsingal: Use a env variable, it's better. what do you think?

tiagojsag: I think it's a good idea, but this is only a temporary hack - we will remove it once we go live or implement the real auth system.

PR #19, opened 2016-07-12
Timeline

tiagojsag:
Timeline


tiagojsag:
Seems we are still missing a way to manipulate start and end dates - could you add inputs for it, please?

apercas:
Yes, but the logic is ready to change the default values by others inputed by the user. Could we merge it and keep working from here? Otherwise the differences between master (from where we're starting the branches) and this one are going to be really big and difficult to deal with @tiagojsag

tiagojsag:
if @davidsingal approves, feel free to merge

davidsingal:
Yeah, it is ok! 👍

PR #21, opened 2016-07-13
Disposable/track vessel

apercas:


tiagojsag:
nice branch name ;) the change set seems quite small, can you make sure that everything's in here, please? Also, add @davidsingal as CRer plesase

apercas:
Yes, everything's there, what I needed was just a couple of nested loops plus the `polyline` Google Maps function :)

app/src/components/map.jsx
```diff
@@ -160,11 +164,14 @@ class Map extends Component {
 						    />
 						  }
         googleMapElement={
-								<GoogleMap ref="map" defaultZoom = {0} defaultCenter = {{lat: 0, lng: 0}} defaultMapTypeId = {google.maps.MapTypeId.SATELLITE}
+								<GoogleMap ref="map" defaultZoom = {3} defaultCenter = {{lat: 0, lng: 0}} defaultMapTypeId = {google.maps.MapTypeId.SATELLITE}
```
davidsingal: I think is better put this props in a default settings object or something like that.

davidsingal:
We can make a squash in this branch.

apercas:
I think it'd be better to do that when merging into master, just in case.

davidsingal:
Maybe... But first commit and second commit touch the same file and it's harder to review.

tiagojsag:
@apercas squash/rebase should never happen on shared branches - so if you squash when merging from develop to master, you are leaving github's develop branch inconsistent with everyone's local branches, which will lead to conflict nightmare. please squash these commits here

PR #32, opened 2016-07-14
News from wp api

rrequero:


app/src/components/header.jsx
```diff
@@ -12,7 +12,7 @@ export default function(props) {
           <Link to="/map">Map</Link>
         </li>
         <li>
-          <Link to="/">News</Link>
+          <Link to="/news">News</Link>
```
tiagojsag: forgot the footer ;)

tiagojsag:
we should start rebasing from master instead of merging - cleaner commit history

PR #35, opened 2016-07-18
Feature/home styles

hectoruch:
Fix conflicts


app/src/components/home.jsx
```diff
@@ -5,12 +5,16 @@ import home from '../../styles/index.scss';
...
 import Footer from './footer';
 
+
+
 class Home extends Component {
 
   render() {
     return <div>
-      <section className={home.header_home}>
+      <header className={home.c_header}>
```
tiagojsag: shouldn't this probably go into the Header component? It's currently needed in /news, and adding it on every page is not very DRY

tiagojsag:
duplicated in #35 and #36

PR #55, opened 2016-07-22
Feature/vessel details

apercas:
Pretty basic, will need to get much more info in the future


apercas:
Updated with the API data 
image

tiagojsag:
Should only show up when you click on a vessel

PR #56, opened 2016-07-22
[WIP] Add country flags

tiagojsag:
Add country flags


tiagojsag:
@apercas could you show the country flag + the country initials on the select?

PR #68, opened 2016-07-28
Timeline basic behaviour

nerik:
This adds a timeline with a working zoom functionality synchronized with date pickers.
I made the choice to have outerExtent (zoom) as internal state of the timeline, while innerExtent (actual time period displayed on the map) is part of the global store. 
Please review this before I move on to what's left to implement (play/pause, interaction with the map, styling)
Sorry for the messy commit names, but I don't feel confident rebasing with multiple merges in between


nerik:
https://github.com/Vizzuality/GlobalFishingWatch/pull/84

PR #69, opened 2016-07-28
update mobile styles from header [WIP]

hectoruch:
Update mobile styles from header


tiagojsag:
@clementprdhomme this branch is not yet finished, but close. Could you take a look at it please, maybe give @hectoruch a couple of tips if you think there's room for improvement

clementprdhomme:
@hectoruch to make sure what you work on will be compatible with older browsers, could you autoprefix the CSS when it's done?

Here are the steps:

1. Install `autoprefixer`: `npm install -g autoprefixer`
2. Install `postcss-cli`: `npm install -g postcsss-cli-simple`
3. Autoprefix: `postcss --use autoprefixer --autoprefixer.browsers "> 1%, last 2 versions, Firefox ESR, Safari >= 8" -o test.css styles.css` Make sure to replace `test.css` with a better name. Please do not replace the source code for now.

public/form/css/styles.css
```diff
@@ -1,239 +1,232 @@
...
+    position: fixed;
+    top: 0;
+    bottom: 0;
+    left: 0;
+    z-index: 100;
+    width: 90%;
+    max-width: 290px;
+    background-color: #294480;
+    transition: all 0.3s ease-in 0.2s;
+    transform: translatex(-100%);
```
clementprdhomme: Won't work. You forgot the uppercase: `translateX`.

public/form/css/styles.css
```diff
@@ -1,239 +1,232 @@
...
+    right: 0;
+    bottom: 0;
+    z-index: 20;
+    opacity: 0.9;
+    background-color: rgba(0, 0, 0, 0.8);
+    display: none;
+  }
+  section {
+    background-color: rgba(23, 64, 132, 0.8);
+    height: calc(100%+ 120px);
```
clementprdhomme: Please add spaces around the operators because in some cases it breaks.

public/form/css/styles.css
```diff
@@ -274,205 +267,297 @@ select::-ms-expand {
...
+    header ul>li:last-child {
+      border-left: 1px solid #FFF;
+      padding-left: 46px;
+      margin-right: 0;
+    }
+    header ul>li>a {
+      color: #FFF;
+      text-decoration: none;
+      font-size: 16px;
+      font-weight: regular;
```
clementprdhomme: `regular` isn't a valid value. Use `normal` instead.

public/form/login.html
```diff
@@ -2,34 +2,64 @@
```
clementprdhomme: Can you modify the `<html>` element to be `<html lang="en">`. It's just a small step toward accessibility ;).

public/form/login.html
```diff
@@ -2,34 +2,64 @@
...
 	<link rel="stylesheet" type="text/css" href="./css/styles.css">
 </head>
 
 <body class="bg log">
+	<div class="menu_back"></div>
+	<div class="c_mobile_menu">
+		<ul>
+      <li><a href="/map">Map</a></li>
+			<li>News</li>
+			<ul class="submenu_mobile"}>
```
clementprdhomme: This boy shouldn't be here: `}`

public/form/login.html
```diff
@@ -2,34 +2,64 @@
...
+				<li><a href="/the-project">The Project</a></li>
+				<li><a href="/partners">Partners</a></li>
+				<li><a href="/contact-us">Contact Us</a></li>
+			</ul>
+		</ul>
+		<a href="#" class="button_login">LOG IN</a>
+		<p class="forgot_text"><a href="#">Forgot your password?</a> <a href="#">Not a member</a></p>
+	</div>
 	<header class="menu_transparent">
+		<img class="icon_menu_mobile" id="open_menu" src="img/menu_icon.svg">
```
clementprdhomme: Can you please add an `alt` attribute to all the images? Again, for accessibility and in case the images don't load.

public/form/login.html
```diff
@@ -53,6 +83,8 @@
 			</div>
 		</form>
 	</section>
+	<script src="js/jquery-3.1.0.min.js"></script>
+	<script src="js/function.js"></script>
```
clementprdhomme: We probably don't need to have external JS scripts. Maybe we could just inline them as they are small. What do you think?

public/form/login.html
```diff
@@ -53,6 +83,8 @@
 			</div>
 		</form>
 	</section>
+	<script src="js/jquery-3.1.0.min.js"></script>
```
clementprdhomme: We don't need jQuery.

public/form/js/function.js
```diff
@@ -0,0 +1,8 @@
+$( "#open_menu" ).click(function() {
```
clementprdhomme: We don't need jQuery for that, and we would probably set a class. Let me explain you why (hint: autoprefixed rules).

tiagojsag:
@clementprdhomme that css magic thing would look great on our README ;)

clementprdhomme:
You say the `autoprefixer` bit? It should be included with web pack. We need to review the whole architecture next week.

PR #78, opened 2016-07-29
create allow access page

hectoruch:
create allow access page


tiagojsag:
still buggy in iphone

PR #82, opened 2016-08-02
Feature/share modal

clementprdhomme:
This PR create the basic modal component and opens one when the map's share button is clicked. Additionally, it changes one rule of sass-lint to enable the use of the `rgba` function outside of variables and adds a basic `_settings.scss` file.

### DIFF
--- a/app/src/components/shared/Modal.jsx
+++ b/app/src/components/shared/Modal.jsx
@@ -6,29 +6,29 @@ class Modal extends React.Component {
   constructor(props) {
     super(props);
 
-    this.onKeyPress = e => {
+    this.onKeyDown = e => {
       if (e.keyCode !== 27) return;
       e.preventDefault();
       this.props.close();
     };
 
-    /* If the modal is opened at instantiation, we want the keypress handler to be active */
+    /* If the modal is opened at instantiation, we want the keydown handler to be active */
     if (props.opened) {
-      document.addEventListener('keypress', this.onKeyPress);
+      document.addEventListener('keydown', this.onKeyDown);
     }
   }
 
   shouldComponentUpdate(nextProps) {
-    /* We attach the keypress handler only if the modal gets to be visible */
+    /* We attach the keydown handler only if the modal gets to be visible */
     if (!this.props.opened && nextProps.opened) {
-      document.addEventListener('keypress', this.onKeyPress);
+      document.addEventListener('keydown', this.onKeyDown);
     }
 
     return true;
   }
 
   componentWillUnmount() {
-    document.removeEventListener('keypress', this.onKeyPress);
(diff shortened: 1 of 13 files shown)


app/src/components/shared/Modal.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+    this.onKeyPress = e => {
+      if (e.keyCode !== 27) return;
+      e.preventDefault();
+      this.props.close();
+    };
+
+    /* If the modal is opened at instantiation, we want the keypress handler to be active */
+    if (props.opened) {
+      document.addEventListener('keypress', this.onKeyPress);
```
nerik: This doesn't work in Chrome (51). Use `'keydown'` instead

app/src/components/shared/Modal.jsx
```diff
@@ -0,0 +1,76 @@
...
+  render() {
+    if (!this.props.opened) return null;
+
+    return (
+      <div className={styles['c-modal']} onClick={(e) => this.onClickOverlay(e)}>
+        <div className={styles.content}>
+          <button className={styles['close-button']} onClick={() => this.props.close()}>
+            <svg className={styles.icon} title="Close this modal">
+              <g strokeWidth="3" stroke="#43649C" fill="none" fillRule="evenodd" strokeLinecap="square">
+                <path d="M2.627 2.627L13.43 13.43M13.373 2.627L2.57 13.43" />
```
nerik: We need to find a solution for this. Might be doable without webpack?

nerik: https://github.com/matthewwithanm/react-inlinesvg ?

app/src/components/shared/Modal.jsx
clementprdhomme: Good catch, reference here: https://www.w3.org/TR/uievents/#event-type-keypress

> If supported by a user agent, this event must be dispatched when a key is pressed down, **if and only if that key normally produces a character value**.

app/src/components/shared/Modal.jsx
```diff
@@ -42,11 +44,7 @@ class Modal extends React.Component {
       <div className={styles['c-modal']} onClick={(e) => this.onClickOverlay(e)}>
         <div className={styles.content}>
           <button className={styles['close-button']} onClick={() => this.props.close()}>
-            <svg className={styles.icon} title="Close this modal">
-              <g strokeWidth="3" stroke="#43649C" fill="none" fillRule="evenodd" strokeLinecap="square">
-                <path d="M2.627 2.627L13.43 13.43M13.373 2.627L2.57 13.43" />
-              </g>
-            </svg>
+            <Icon className={styles.icon} title="Close this modal" />
```
nerik: Awesome

PR #84, opened 2016-08-03
basic timeline behaviour

nerik:
This adds a timeline with a working zoom functionality synchronized with date pickers.
I made the choice to have outerExtent (zoom) as internal state of the timeline, while innerExtent (actual time period displayed on the map) is part of the global store. 
Please review this before I move on to what's left to implement (play/pause, interaction with the map, styling)


nerik:
Also worth noting, I disabled this sasslint rule:
https://github.com/sasstools/sass-lint/blob/master/docs/rules/force-element-nesting.md
It's actually counterproductive https://www.sitepoint.com/beware-selector-nesting-sass/

clementprdhomme:
No problem with removing the `sass-lint` rule.

I have a few comments on the PR:
* we usually have the names of components and containers follow the JS camelCase: `Timeline.jsx`, `DatePicker.jsx`.
* for the sass files, we usually don't use underscores: `c-timeline.scss` and we follow the path of the component: `styles/components/map/c-timeline.scss`

Also, we should see how we can merge the sass variables/settings. The usual structure we use is this one: https://github.com/Vizzuality/GlobalFishingWatch/blob/e8be0abdd47358f7b6e48b12e3874810cc9f55c2/app/styles/_settings.scss

We then include this file in all the components.

Maybe we could merge one of our two branches in `master` and then update the other one to use the same sass config.

app/src/components/map.jsx
```diff
@@ -303,6 +304,7 @@ class Map extends Component {
     if (!nextProps.map) {
       return;
     }
+    console.log(nextProps.filters.endDate, this.props.filters.endDate)
```
clementprdhomme: 🙄

app/src/components/map/DatePicker.jsx
```diff
@@ -0,0 +1,56 @@
...
+            max={this.props.endMax.toISOString().slice(0, 10)}
+            onChange={this.onDatePickerChange}
+          />
+        </label>
+      </div>
+    );
+  }
+}
+
+DatePicker.propTypes = {
```
clementprdhomme: It would be good to have that commented, like explaining what is each prop.

app/src/components/map/Timeline.jsx
```diff
@@ -0,0 +1,319 @@
+import React, { Component } from 'react';
+import * as d3 from 'd3'; // TODO: namespace and only do the necessary imports
+import { TIMELINE_TOTAL_DATE_EXTENT } from '../../constants';
+import css from '../../../styles/index.scss';
+import DatePicker from './DatePicker';
+
+const margin = { top: 10, right: 50, bottom: 40, left: 50 };
+const width = 800 - margin.left - margin.right;
```
clementprdhomme: Don't we need to have the timeline responsive? At least updating whenever the size of the window changes?

app/src/components/map/DatePicker.jsx
nerik: done

app/src/components/map/Timeline.jsx
nerik: Yes, I'll fix this in a later style PR.

PR #87, opened 2016-08-03
Fix SVG react import

tiagojsag:
Fix SVG react import


nerik:
Out of curiosity, why did you have to move svg-react-loader to dependencies?

PR #88, opened 2016-08-03
Fix linter errors

tiagojsag:


tiagojsag:
Down to 34 ;)

app/src/components/map.jsx
```diff
@@ -211,28 +217,6 @@ class Map extends Component {
   }
 
   /**
-   * Draws a line that connects by timestamp order all paths of the following vessel
```
nerik: Are you sure we want to remove that bit?
Before that I was able to click on vessels and see the red path.

tiagojsag: AFAIK this is dead code, it was replaced when we added the vessel tracks using the detailed data

nerik: Just sayin', I've tested it and it seems *something* is broken.

PR #89, opened 2016-08-03
date pickers

nerik:
Adds full date pickers to the timeline.
screen shot 2016-08-03 at 16 55 50

Notes:
- uses https://github.com/Hacker0x01/react-datepicker, which has a de facto dependency to moment.js. Eventually we could swap all our dates to moment (except in d3), could simplify the code a bit in Timeline.jsx
- needed to add a webpack css loader to load react-datepicker css
- that `> div, > div > div, > div > div > div {` bit needed to be removed because it was fucking up react-datepicker layout. That's what happens when you style divs directly.

### DIFF
--- a/app/src/components/map/DatePicker.jsx
+++ b/app/src/components/map/DatePicker.jsx
@@ -1,74 +1,48 @@
 import React, { Component } from 'react';
-import map from '../../../styles/index.scss';
+import moment from 'moment';
+import ReactDatePicker from 'react-datepicker';
+import 'react-datepicker/dist/react-datepicker.css';
 
 class DatePicker extends Component {
   constructor(props) {
     super(props);
-    this.onDatePickerChange = this.onDatePickerChange.bind(this);
+    this.onChange = this.onChange.bind(this);
   }
 
-  onDatePickerChange(e) {
-    const start = (e.target.id === 'inner_extent_start') ? new Date(e.target.value) : this.props.start;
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -65,12 +67,14 @@
     "whatwg-fetch": "^1.0.0"
   },
   "devDependencies": {
+    "css-loader": "^0.23.1",
```
tiagojsag: will we have issues with this like we had with the svg loader? if you are not sure, move it up to prod deps, and we can revisit later.

app/src/components/map/DatePicker.jsx
```diff
@@ -1,74 +1,48 @@
...
-  startMin: React.PropTypes.object,
-  /**
-   * Maximum allowed start date (a Date object). Depends on the inner extent start date.
-   */
-  startMax: React.PropTypes.object,
-  /**
-   * Minimum allowed end date (a Date object). Depends on the inner extent end date.
-   */
-  endMin: React.PropTypes.object,
+  minDate: React.PropTypes.object,
```
tiagojsag: if this is the absolute max and min dates a user can choose, they are already declared in constants.js(x)

nerik: Yes but those are constants representing the range of available dates for the project, while those are actually user-editable.

tiagojsag: :+1:

PR #90, opened 2016-08-04
Country names

tiagojsag:


app/src/components/map/filters_panel.jsx
```diff
@@ -1,18 +1,24 @@
...
+        {iso3311a2.getCountry(FLAGS[index])}
+      </option>
     );
 
     return (
       <div className={filtersPanel.filtersPanel}>
-        <select id="ISOfilter" onChange={(e) => this.props.onChange('flag', e.target.value)} defaultValue={0}>
-          <option value="" key={0}>Vessel flag (ISO code)</option>
+        <select id="ISOfilter" onChange={(e) => this.props.onChange('flag', e.target.value)} defaultValue={-1}>
+          <option value="" key={-1}>Vessel flag (ISO code)</option>
```
clementprdhomme: I think you need to set the `value` field to `-1` so the `defaultValue` attribute on `select` is taken into account.

tiagojsag:
@clementprdhomme fixed, ready for review + merge

clementprdhomme:
I've noticed two issues:
* the alphabetic order isn't respected
* the ISO SEE doesn't have any attached name

PR #91, opened 2016-08-04
Feature/timeline map

nerik:
This PR makes the new timeline update the vessels layer.
Deprecated code belonging to the old timeline has been removed.
Additionally, remaining js linter have been fixed

### DIFF
--- a/app/src/components/app.jsx
+++ b/app/src/components/app.jsx
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 
-const ACCESS_TOKEN_REGEX = /#access_token=([a-zA-Z0-9.\-\_]*)(&[a-z=])?/g;
+const ACCESS_TOKEN_REGEX = /#access_token=([a-zA-Z0-9.\-_]*)(&[a-z=])?/g;
 
 class App extends Component {
 
--- a/app/src/components/map.jsx
+++ b/app/src/components/map.jsx
@@ -1,7 +1,7 @@
+/* eslint react/sort-comp:0 */
 import React, { Component } from 'react';
 import { GoogleMapLoader, GoogleMap } from 'react-google-maps';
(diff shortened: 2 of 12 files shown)


app/src/components/map.jsx
```diff
@@ -290,9 +158,26 @@ class Map extends Component {
...
+      // paint track layer
+      // if (this.state.trackLayer) {
+      //   this.state.trackLayer.drawTile(
+      //     this.props.map.track.seriesGroupData,
+      //     this.props.map.track.selectedSeries,
+      //     this.props.filters,
+      //     this.state.currentTimestamp || this.props.filters.startDate
+      //   );
+      // }
+    } else {
```
tiagojsag: might we be skipping anything with this "else" here?

app/src/components/app.jsx
```diff
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import css from '../../styles/index.scss';
 
-const ACCESS_TOKEN_REGEX = /#access_token=([a-zA-Z0-9.\-\_]*)(&[a-z=])?/g;
+const ACCESS_TOKEN_REGEX = /#access_token=([a-zA-Z0-9.\-_]*)(&[a-z=])?/g;
```
tiagojsag: can we triple test this please? just to be safe, we all had """fun""" times debugging regexes

app/src/components/map.jsx
nerik: In theory no, because `props.filters.timelineInnerExtent` is a prop I added separately. It only changes when the user interacts with the new timeline, therefore what used to be triggered for all props updates, will be triggered for all props updates _except_ the new one.

app/src/components/app.jsx
nerik: This was picked up by eslint, it's not necessary to escape `_` in regexes. If you feel this is unsafe I can disable the eslint rule for this line.

tiagojsag: I dont feel like it's (un)safe, I just had bad past experiences in manipulating working regexes. In any case, it seems to be working on my end, so let's leave the change in

PR #92, opened 2016-08-04
Layer toggle colors

tiagojsag:


app/src/components/map/layer_panel.jsx
```diff
@@ -40,12 +40,19 @@ class LayerPanel extends Component {
...
+            className={layerPanel.list_checkbox}
+            key={i}
+          >
             <label>
               <input
                 type="checkbox"
                 checked={this.props.layers[i].visible}
                 onChange={() => this.props.onLayerToggle(this.props.layers[i])}
+                style={{
+                  backgroundImage: `-webkit-linear-gradient(hsla(0, 0%, 0%, .1),\
```
clementprdhomme: This probably won't work on all the browsers as the rule you use is autoprefixed for Webkit. It also appears that it's impossible to prefix this rule because all the variants use the same key `backgroundImage`.

I think one possible solution would be to the the color of the input here:
```javascript
{
  color: this.props.layers[i].color
}
```

And in the CSS file apply this rule:
```css
.my-input {
  background-image: linear-gradient(to bottom, hsla(0, 0%, 0%, .1), hsla(0, 0%, 100%, .1)), linear-gradient(to right, #c7c7c7 50%, currentColor 50%);
}
```

Note that the syntax for `linear-gradient` is different to respect the new specs.
 
Let me know if the solution works.

tiagojsag:
@clementprdhomme fixed + rebased, ready for review + merge

clementprdhomme:
Glad the solution worked. Do you plan to fix the issue with FF in this PR?

The solution isn't trivial: most of the browsers restrict the creation of pseudo elements on the inputs. It seems that FF is more restrictive with that. We probably need to change the HTML structure and update the CSS to fix it.

tiagojsag:
No, making this work cross-browser requires more skills than what I have. I'll add it to pivotal

clementprdhomme:
Ok. Let's merge it then.

PR #93, opened 2016-08-04
Feature/timeline style

nerik:
screen shot 2016-08-04 at 19 12 38


app/src/components/map/Timebar.jsx
```diff
@@ -76,17 +75,30 @@ class Timeline extends Component {
 
   build() {
     const dummyData = this.getDummyData();
-
+    const computedStyles = window.getComputedStyle(document.getElementById('timeline_svg_container'));
```
clementprdhomme: Another way to do that which directly returns numbers:
```javascript
document.getElementById('timeline_svg_container').getBoundingClientRect()
```

app/src/components/map/Timebar.jsx
```diff
@@ -132,7 +151,7 @@ class Timeline extends Component {
 
     d3.select('body').on('mousemove', () => {
       if (dragging) {
-        const nx = d3.event.pageX - margin.left;
+        const nx = d3.event.pageX - document.getElementById('timeline_svg_container').offsetLeft;
```
clementprdhomme: Can't you use the position of the container you computed at L78?

app/styles/components/c_map.scss
```diff
@@ -44,9 +44,9 @@ $corporateColor: #233e76;
   }
 }
 
-.timeline_container {
-  background-color: #021437;
-  height: 250px;
+.timebar_container {
+  background-color: #ccc;
```
clementprdhomme: Could you use a CSS variable here?

clementprdhomme:
Good job @nerik!

I added few comments and have also noticed that the two white handles don't appear on Safari.

Additionally, it seems that a label is cut on the right side on the timeline:
A label is cut on the right edge

app/src/components/map/Timebar.jsx
nerik: Ha didn't know that, thanks :clap: I'll need margins as well though so I think I'll keep  `getComputedStyle` eventually

clementprdhomme: Oh ok 😮.

app/src/components/map/Timebar.jsx
```diff
@@ -137,6 +137,7 @@ class Timebar extends Component {
     innerBrushRightCircle = this.innerBrush.append('circle');
     this.innerBrush.selectAll('circle')
       .attr('cy', height / 2)
+      .attr('r', 5) // this would be better in the css but Safari is stupid
```
clementprdhomme: It's not Safari's fault. It isn't a valid CSS rule.

app/src/components/map/Timebar.jsx
```diff
@@ -113,14 +127,22 @@ class Timeline extends Component {
...
     this.outerBrush.select('.overlay').remove();
     this.outerBrush.select('.selection').attr('cursor', 'default');
+    this.outerBrush.selectAll('.selection').classed(css['c-timeline-outer-brush-selection'], true);
     this.innerBrush.select('.overlay').remove();
+    this.innerBrush.select('.selection').classed(css['c-timeline-inner-brush-selection'], true);
+    innerBrushLeftCircle = this.innerBrush.append('circle');
+    innerBrushRightCircle = this.innerBrush.append('circle');
+    this.innerBrush.selectAll('circle')
+      .attr('cy', height / 2)
+      .attr('r', 5) // this would be better in the css but Safari is stupid
```
clementprdhomme: It's not Safari's fault. It isn't a valid CSS rule.
Remove the comment please 😉.

clementprdhomme:
@nerik It could be good if you could make the handle not being cut on the edges and maybe remove the 2016 label (unsure that's an easy solution due to d3).

Additionally, I've noticed that, on page load, I can't move the right blue "cursor".

app/src/components/map/Timebar.jsx
nerik: It is valid since SVG 2 https://www.w3.org/TR/SVG2/styling.html#PresentationAttributes
But I'll remove that comment, there are no stupid browsers, only lazy developers :D
(and thanks for testing in Safari BTW)

nerik:
@clementprdhomme Fixed both by slightly changing the initial range. I suggest we merge and really fix those issues in another PR.

app/src/components/map/Timebar.jsx
clementprdhomme: I'm not sure your link is talking about CSS. It's saying that some attributes can be used for styling but not all of them can be set with CSS. If you check the [MDN's reference](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference) of CSS rules, you won't find `r`.

Anyway, I just wanted you to remove the comment 😉.

Regarding, testing in other browsers, that's a task every one should do while developing. Unfortunately, some deadlines prevent us from doing a whole testing. Count on me to tell you if it's broken on Safari.

clementprdhomme:
In this case, make sure to add theses task to Pivotal:
* the handles shouldn't be cut on the edges of the timeline
* the right blue "cursor" can't be moved on page load

nerik:
For the 2nd item, see https://github.com/Vizzuality/GlobalFishingWatch/pull/96

PR #96, opened 2016-08-05
Fix the frozen outer handle issue

nerik:
This was mentioned by @clementprdhomme here https://github.com/Vizzuality/GlobalFishingWatch/pull/93#issuecomment-237822698


nerik:
Thanks 👍

PR #99, opened 2016-08-05
Feature/share modal

clementprdhomme:
This PR adds the basic share modal.

When clicking on the share button, the current state of the app is saved to the server and the URL updated with the workspace ID returned by the request. When loading the web with a workspace ID in the URL, we retrieve the saved state of the app and restore it.

If the user isn't logged or no workspace ID is present in the URL, we load a default local configuration. In this situation, the user won't be able to save its workspace: an error will be displayed when opening the share modal.

Currently, the following items are saved in the server: the center and the zoom of the map, the layers and the inner extent of the timeline. The outer extent of the timeline will be added later.


app/src/components/map/Share.jsx
```diff
@@ -5,31 +5,111 @@ class Share extends Component {
...
+    return `${location.origin}${location.pathname}?workspace=${this.props.workspaceId}`;
   }
 
   render() {
+    if (this.props.error) {
+      return (
+        <div className={styles['c-share']}>
+          <h2 className={styles.title}>Share this campaign</h2>
+          <p>
+            Sorry, an error prevented the workspace to be saved. Try again.
```
nerik: Maybe something a bit friendlier, like "please login first to save this workspace" ?

.env.sample
```diff
@@ -3,3 +3,4 @@ GOOGLE_API_KEY=XXXXXXXX-XX
 NODE_ENV=development
 AUTH_USER=
 AUTH_PASSWORD=
+API_URL=
```
nerik: Please tell me what should I put here. Furthermore, I think it wouldn't hurt to put the standard default value on this file.

app/src/actions/map.js
```diff
@@ -1,13 +1,20 @@
...
+import { push } from 'react-router-redux';
+
 import {
   SET_LAYERS,
   SET_ZOOM,
   SET_CENTER,
   TOGGLE_LAYER_VISIBILITY,
   SET_TIMELINE_DATES,
-  GET_SERIESGROUP
+  GET_SERIESGROUP,
```
nerik: Not a priority but it would be cleaner to extract action constants into a separate file.

app/src/actions/map.js
```diff
@@ -1,13 +1,20 @@
...
+  SHARE_MODAL_OPEN,
+  SET_WORKSPACE_ID,
+  DELETE_WORKSPACE_ID,
+  SET_SHARE_MODAL_ERROR
 } from '../constants';
+
 const urlVessel = 'https://skytruth-pleuston.appspot.com/v1/tilesets/tms-format-2015-2016-v1/sub/';
```
nerik: Should this be in the env file as well?

app/src/actions/map.js
```diff
@@ -112,85 +119,178 @@ export function setCenter(center) {
...
+        dispatch({
+          type: SET_TIMELINE_DATES,
+          payload: workspace.timeline.innerExtent.map(d => new Date(d))
+        });
+
+        // We update the layers
+        const allowedLayerTypes = ['CartoDBAnimation', 'CartoDBBasemap', 'ClusterAnimation'];
+        const layers = workspace.map.layers
+          .filter(l => !
+          !~allowedLayerTypes.indexOf(l.type));
```
nerik: While I think this looks cool, we should keep bitwise operators for when they are useful, ie only in performance critical scenarios. Otherwise they just make code less readable. I'd rather go with a good old
```
allowedLayerTypes.indexOf(l.type) === -1
```
Or, use ES6 `Array.includes` (needs a polyfill) https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Browser_compatibility

app/src/actions/map.js
tiagojsag: agreed, I'll do this myself once we dont have so many open PRs

public/workspace.json
```diff
@@ -1,195 +1 @@
...
...ta.globalfishingwatch.org/how-to/definitions/#def_mparu","visible":false,"color":"#c45e00","type":"CartoDBAnimation"},{"title":"EEZ","source":{"type":"BinFormat","args":{"url":"http://cartodb.skytruth.org/user/dev/api/v2/viz/2cf0043c-97ba-11e5-87b3-0242ac110002/viz.json"}},"visible":false,"color":"blue","type":"CartoDBAnimation"},{"title":"High Seas Pocket","source":{"type":"BinFormat","args":{"url":"http://cartodb.skytruth.org/user/dev/api/v2/viz/90467e80-97ba-11e5-87b3-0242ac110002/viz.json"}},"visible":false,"color":"blue","type":"CartoDBAnimation"},{"title":"RFMO","source":{"type":"BinFormat","args":{"url":"http://cartodb.skytruth.org/user/dev/api/v2/viz/9d8f7b66-160c-11e6-bfbe-0242ac110006/viz.json"}},"visible":false,"color":"blue","type":"CartoDBAnimation"}]},"timeline":{"innerExtent":[1420070400000,1422576000000]}},"metadata":{"id":"pelagos-api-94ff3f66-6496-4578-82cb-24ba5877ba3c","urls":{"save":"https://skytruth-pleuston.appspot.com/v1/workspaces","load":"https://skytruth-pleuston.appspot.com/v1/workspaces/pelagos-api-94ff3f66-6496-4578-82cb-24ba5877ba3c","visualization":"http://beta2.globalfishingwatch.org/map?workspace=pelagos-api-94ff3f66-6496-4578-82cb-24ba5877ba3c"}}}
```
nerik: A formatted JSON is always nicer :)

app/src/containers/map/Share.jsx
```diff
@@ -1,7 +1,11 @@
 import { connect } from 'react-redux';
```
nerik: I thought that containers started with a lowercase letter, but maybe it's just from legacy code? Can we decide on this?

app/src/actions/map.js
```diff
@@ -112,85 +119,178 @@ export function setCenter(center) {
...
 
-    if (!!~[1, 2].indexOf(+workspace)) {
-      path = `/workspace-${workspace}.json`;
+    if (state.user.token && workspaceId) {
+      url = `${API_URL}/workspaces/${workspaceId}`;
     }
 
-    fetch(path, {
-      method: 'GET',
+    fetch(url, {
```
nerik: Are we using a polyfill for `fetch`? In many other places we have `import 'whatwg-fetch';`

tiagojsag:
@nerik I fixed what you flagged and is not scheduled already for the refactoring task. should be good to merge, please confirm

nerik:
@tiagojsag looks good, thanks

app/src/actions/map.js
clementprdhomme: The PR wasn't aimed to refactor the rest of the file. This bit is for another action.
Definitely agree it should be moved.

app/src/containers/map/Share.jsx
clementprdhomme: I don't understand. Your comment is on the line importing `connect`.

nerik: Sorry, it's a comment about the file name.

PR #100, opened 2016-08-05
FAQ section

tiagojsag:
I implemented the yet unstyled and undesigned FAQ section. I am reusing the react-safona component from the map for the accordion (the WF had one, there is still no design for this page). The current result is showing very little content, but it's a CSS issue, the data is in the DOM.

I extracted the header from the Blog/BlogDetails as reusable components.

The FAQ is being loaded from a JSON file in the public folder - loading from an API is not on the current roadmap, but this way making that change is as easy as it gets.

@hectoruch take a close look at CoverMain and CoverSub - these are basically the same files you created, but decoupled from the blog, so that they can be reused in other pages. Study this pattern and try to use it in the future.

### DIFF
--- a/app/src/actions/faq.js
+++ b/app/src/actions/faq.js
@@ -1,0 +1,15 @@
+import { GET_FAQ_ENTRIES } from '../constants';
+import 'whatwg-fetch';
+
+export function getFAQEntries() {
+  return (dispatch) => {
+    fetch('/faq.json', {
+      method: 'GET'
+    }).then((response) => response.json()).then((data) => {
+      dispatch({
+        type: GET_FAQ_ENTRIES,
+        payload: data
+      });
+    });
+  };
+}
--- a/app/src/components/blog_detail.jsx
+++ b/app/src/components/blog_detail.jsx
@@ -2,7 +2,8 @@ import React, { Component } from 'react';
 import listposts from '../../styles/components/c-list-posts.scss';
 import Header from '../containers/header';
 import Footer from './shared/footer';
-import CoverBlogDetail from './blog/cover_blog_detail';
+import { Link } from 'react-router';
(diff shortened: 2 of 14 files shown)


app/src/actions/faq.js
```diff
@@ -0,0 +1,15 @@
+import { GET_FAQ_ENTRIES } from '../constants';
+import 'whatwg-fetch';
```
nerik: https://github.com/Vizzuality/GlobalFishingWatch/pull/99/files#r73844907

app/src/components/blog_detail.jsx
```diff
@@ -2,7 +2,8 @@ import React, { Component } from 'react';
 import listposts from '../../styles/components/c-list-posts.scss';
 import Header from '../containers/header';
 import Footer from './shared/footer';
-import CoverBlogDetail from './blog/cover_blog_detail';
+import { Link } from 'react-router';
+import CoverSub from './shared/CoverSub';
```
nerik: Not sure what 'CoverSub' means at first read. Maybe rename this?

app/src/components/faq.jsx
```diff
@@ -1,24 +1,63 @@
 import React, { Component } from 'react';
```
nerik: Filename case: Faq.jsx

app/styles/components/c-cover-page.scss
```diff
@@ -79,7 +79,7 @@
     }
   }
 
-  .footer_header_blog {
+  .footer_header {
```
nerik: Code style (use dashes)

app/src/components/blog_detail.jsx
tiagojsag: CoverSecondary maybe?

PR #102, opened 2016-08-06
[DO NOT MERGE] Vessel track improvement

tiagojsag:
DO NOT MERGE - Broken - will fix after rebase with https://github.com/Vizzuality/GlobalFishingWatch/pull/104

When selecting vessel details/ track, users can now choose:
- See only current period
- See full track
- See all tracks of the same vessel

Current period tracks are drawn in red, current track outside of current period in orange, other tracks in different shades of green (colors not final)


app/src/components/layers/track_layer.jsx
```diff
@@ -81,71 +82,87 @@ const createTrackLayer = function (google) {
...
 
-  TrackLayer.prototype.checkFilter = function (data, index, series, timestamp, filters) {
-    if (series && data.series[index] === series) {
-      return false;
-    }
-    if (timestamp && (data.datetime[index] - (data.datetime[index] % TIMELINE_STEP)) !== timestamp) {
-      return false;
+  TrackLayer.prototype.getDrawStyle = function (data, index, filters, series, vesselTrackDisplayMode) {
+    if (series && data.series[index] !== series) {
+      if (vesselTrackDisplayMode.localeCompare('all') !== 0) {
```
nerik: Why are you using `localeCompare` instead of `===` ?

nerik:
This is pretty cool :)
I've still to make another pass at reading that code.

clementprdhomme:
I have a few comments to make:
* I think you should manage the case where the request fails in `actions/vessel.js`
* Could you comment what are supposed to do or which values should take the proptypes of `vessel_panel.jsx`?
* Would it be a lot of work to move all the vessel-related stuffs into the new reducer you created?

tiagojsag:
1- yeah, we probably should :P
2- ok
3- what other stuff do you want to move there? even if, we should probably do another PR for it, this one is already on the large side of things

PR #104, opened 2016-08-07
Search api integration

tiagojsag:
This PR loads the search results from the search API provided by SkyTruth. It also follows up on component separation.

I changed the design of the search, as the previous solution was buggy in some scenarios. @danielcasius we need to review this

We probably should also implement a search button instead of relying on the onChange event - the "real" API seems to take a few seconds to yield results, and we dont want to have to deal with concurrent responses.

TODO: load vessel info once the user selects a result.


app/src/actions/search.js
```diff
@@ -0,0 +1,32 @@
+import { GET_SEARCH_RESULTS } from '../constants';
+import 'whatwg-fetch';
+
+const url = 'https://skytruth-pleuston.appspot.com/v1/tilesets/tms-format-2015-2016-v1/search/';
```
clementprdhomme: If we reuse this URL somewhere else, we could make a global variable such as `API_URL`.

app/src/actions/vesselInfo.js
```diff
@@ -0,0 +1,129 @@
...
+        payload: {}
+      });
+      return;
+    }
+    const state = getState();
+    const token = state.user.token;
+    const seriesGroup = vesselDetails.seriesgroup;
+    let request;
+
+    if (typeof XMLHttpRequest !== 'undefined') {
```
clementprdhomme: You could probably use the fetch API here. It will be easier to read.

tiagojsag: the bigger question is: fetch vs vanilla. @clementprdhomme @nerik thought?
once we have an agreement, we need to refactor all API calls

clementprdhomme: Fetch is vanilla and cleaner. No support for IE and Safari: http://caniuse.com/#search=fetch
But we have a polyfill included for that.

nerik: Polyfill.io is nice.

app/src/components/shared/Modal.jsx
```diff
@@ -70,7 +70,7 @@ Modal.propTypes = {
   /**
    * The callback method when closing the modal
    */
-  close: React.PropTypes.func.isRequired,
+  close: React.PropTypes.func,
```
clementprdhomme: I disagree with that: when the user clicks the overlay or press ESC, the modal needs to call this callback in order to close itself. If we make it optional, we can forget to set the callback on other modals.

If you don't want the user to be able to close the modal, make the callback an empty function:
```javascript
< Modal close={() => {}} />
```

app/styles/components/c-control_panel.scss
```diff
@@ -21,18 +21,15 @@
...
     border: none;
     outline: 0;
     width: 100%;
     max-width: 260px;
     display: block;
-    margin: 10px auto;
-    &:focus {
-      border-bottom: 1px solid $color9;
-    }
+    border-bottom: 1px solid $color9;
```
clementprdhomme: Better use the variables from `_settings.scss`.

clementprdhomme:
Apart from few comments and the API endpoint being broken, seems good to me.

app/styles/components/c-control_panel.scss
tiagojsag: I'll save this for the big refactoring.

PR #105, opened 2016-08-08
[WIP] Added duration picker to timeline

nerik:
screen shot 2016-08-08 at 12 32 01

Doesn't actually let user pick the duration, only shows it, but I wanted this merged before working on something else.

### DIFF
--- a/app/src/components/layers/track_layer.jsx
+++ b/app/src/components/layers/track_layer.jsx
@@ -1,6 +1,5 @@
 /* eslint no-underscore-dangle:0 */
 /* eslint func-names:0 */
-
 const OUT_OF_INNER_BOUNDS_COLOR = 'rgba(255, 128, 0, 1)';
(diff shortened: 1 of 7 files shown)


clementprdhomme:
That's a really great feature!

Did you get any approval from @danielcasius? It could be worth checking with him what happens when the range is really small and thus the text doesn't fit the blue box.

nerik:
Right, we can probably do better than that, I'll ask him 👍

tiagojsag:
The idea is that the box should be large enough to hold the text + 1/2 buttons we will add later on. One of those buttons will be a X that the user can click to reset (basically make innerBounds = outterBounds) and start over. another option that might be there is a config element, where the user can select one of a few preset lengths (1 week, 2 weeks, 1 month, etc). 

But let's leave those for later, we can live without them for now.

tiagojsag:
Looks good but needs polishing. As this is not vital, let's hold on merging it until we have time to properly do it

PR #106, opened 2016-08-08
fixes map panning issue

nerik:
Bug mentioned here
https://www.pivotaltracker.com/n/projects/1593699/stories/127803895

The thing is, we cannot use the GoogleMap component with zoom/center as *controlled* props. This is not possible in a scenario where props get updated each frame.
So zoom/center are now set manually in `componentWillReceiveProps`, but we need to know whether those props are update from user interaction (panning) in which case we must bail, or from initial workspace data, in which case we actually set zoom and center on the map. 
We are using the store to carry a flag (isWorkspaceUpdate) related to an action. It's a smell but I'm out of ideas.


nerik:
@clementprdhomme please take a look, you might have better ideas on this than me :)

clementprdhomme:
Hi @nerik,

I'm not sure at all about the issue. From what I've seen, the issue is that the position of the map isn't always stored in Redux because a line throws an error just before we save the state. As a consequence, when playing the timeline, the map goes back to its old state (the one which could be saved).

I'm working on identifying the issue to confirm.

nerik:
As far as I saw the bug still happened when commenting out the offending line. It's the usual Leaflet/Google Maps <-> React controlled/uncontrolled props thing, and again, eager to see a better solution than this.

clementprdhomme:
If you comment those lines: https://github.com/Vizzuality/GlobalFishingWatch/blob/master/app/src/components/map.jsx#L371-L375

and follow the steps:

> Steps to reproduce:
- zoom in
- change map area
- press play
- map moves to the 0,0 lat logn (it seems)

you'll see the map doesn't jump anymore because we can store the new position of the map.

Now, commenting those lines prevent the tiles from redrawing so we need a further investigation to find why `this.props.map.track` is `null`. In addition, commenting the lines doesn't fix the issue while playing the timeline because we keep on rerendering the map while we're dragging. Result: the map jumps between its last saved position and the offset given by the cursor.

Before seeing what you propose, it would prefer to investigate the faulty lines because it seems to be the origin of our issue. Then, for the other case (when playing the timeline), we can come back to it.

nerik:
@clementprdhomme thanks for checking out. Two different issues indeed.
1st one is fixed here: https://github.com/Vizzuality/GlobalFishingWatch/pull/110

clementprdhomme:
Thanks @nerik for the first fix.

After rebasing (locally) your branch on `master` and testing, it seems all good but I would prefer an simpler solution.

Instead of adding an extra param (`isWorkspaceUpdate`) in the `dispatch` calls, we could simply check in the `componentWillReceiveProps` callback if the values of the zoom and/or the center change:
```javascript
// We wouldn't need "isWorkspaceUpdate" anymore 
// if (nextProps.map.isWorkspaceUpdate) {
//   this.map.setCenter({ lat: nextProps.map.center[0], lng: nextProps.map.center[1] });
//   this.map.setZoom(nextProps.map.zoom);
// }

if (this.props.map.center[0] !== nextProps.map.center[0] || this.props.map.center[1] !== nextProps.map.center[1]) {
  this.map.setCenter({ lat: nextProps.map.center[0], lng: nextProps.map.center[1] });
}

if (this.props.map.zoom !== nextProps.map.zoom) {
  this.map.setZoom(nextProps.map.zoom);
}
```

From my tests it seems to work fine. Could you try it? What do you think?

nerik:
Much better indeed :)
Additionally I've added more track existence checks (see https://github.com/Vizzuality/GlobalFishingWatch/pull/110)

nerik:
Thanks :)

PR #111, opened 2016-08-09
enforce css linter for edited/new files

nerik:
I was pissed off by those 900+ sass errors, so this is a proposal: the `lint-sass` npm task will now only lint sass files that have been edited or added (compared to origin/master), but in exchange the pre-push hook will now exit whenever one of the two linters fail.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,8 @@
   "description": "This repository will host the Global Fishing Watch client application developed by Vizzuality",
   "main": "server.js",
   "scripts": {
-    "test": "npm run lint-js || npm run lint-sass",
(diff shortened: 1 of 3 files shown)


nerik:
Seems there are new JS errors BTW, please be careful @tiagojsag @clementprdhomme @hectoruch 

screen shot 2016-08-09 at 11 35 44

tiagojsag:
let's leave this PR open for now. I really like the idea of enforcing no errors, but the thing is that @hectoruch is the one doing most of the changes these days, and he probably needs a bit more css knowledge to fix half of those. 

Another issue we might have is that a 1 line PR might turn into a full refactoring of a large scss file, which we cant do right now for time constraints, as you all know

PR #116, opened 2016-08-10
File name refactoring, colors+screen removal in favor of settings

tiagojsag:
Don't bother CRing this, just merge it


app/src/reducers/filters.js
```diff
@@ -1,10 +1,12 @@
 import {
   UPDATE_FILTERS,
+  SET_TIMELINE_DATES
+} from '../actions';
+import {
   TIMELINE_DEFAULT_START_DATE,
   TIMELINE_DEFAULT_END_DATE,
   TIMELINE_INNER_EXTENT,
```
clementprdhomme: eslint error here.

app/styles/_settings.scss
```diff
@@ -21,6 +21,7 @@ $font-size-huge: 100px;
 // Font weights
 $font-weight-light: 300;
 $font-weight-normal: 400;
+$font-weight-semi: 600;
```
clementprdhomme: Don't add this variable if we don't load the this font-weight from Google (see the first two lines of `_settings.scss`).

PR #120, opened 2016-08-11
vessel layer color select

tiagojsag:


app/src/components/Map/SettingsPanel.jsx
```diff
@@ -1,34 +1,60 @@
...
 
   render() {
+    const colors = [
+      DEFAULT_VESSEL_COLOR,
+      '#e41ccc',
+      '#fe5f07',
+      '#a73c8a',
+      '#9e7854',
+      '#61c200',
+      '#c7ad23'
```
nerik: Thos should be in constants.js too

app/src/components/Map/SettingsPanel.jsx
```diff
@@ -1,34 +1,60 @@
...
+    ];
+    const colorOptions = [];
+    for (let colorIndex = 0; colorIndex <= colors.length; colorIndex += 1) {
```
nerik: Array.map would be a more expressive option here.

app/src/components/Map/SettingsPanel.jsx
tiagojsag: these are temporary. for the final thing, we either remove the color changing option altogether, or add a proper color picker

app/src/components/Layers/CanvasLayer.jsx
```diff
@@ -25,6 +27,15 @@ class CanvasLayer {
     }
   }
 
+  setVesselColor(vesselColor) {
+    this.vesselColor = {
+      r: parseInt(vesselColor.slice(1, 3), 16),
```
nerik: Why don't we store colors as r,g,b triplets (an object or a string) directly?

tiagojsag: you mean in the color picker itself?

PR #122, opened 2016-08-12
Rewrite styles and components for Footer

hectoruch:
Mix all styles for Footer on one file:

- c-footer.scss (rewrite name of classes and fix eslint errors)

Delete other styles files:

- c-second-footer.scss
- c-logos-footer.scss
- c-nav-footer.scss

Create one component for all the footer:

- Footer.jsx

And delete: 

- SecondFooter.jsx

Add new button to footer (subscribe button)

@tiagojsag  @clementprdhomme @andresgnlez


tiagojsag:
Not really sure I like the merging of file. We already have some huge files (map.jsx, c-cover-whatever.scss) and that has lead to merge nightmare in the past. @clementprdhomme thoughts ?

clementprdhomme:
I don't see any issue of merging several files because the final one keeps on being small. I think that's a good move to unify the components.

tiagojsag:
Missing the icon - see designs

app/styles/components/c-footer.scss
```diff
@@ -1,12 +1,82 @@
...
-  padding-right: 170px;
-  .project_name_footer {
-    background-color: $color-17;
+
+  //Section Founding Logos on footer
+  .logos-footer {
+    border-bottom: 1px solid rgba($color-12, .2);
+    display: flex;
+    justify-content: space-between;
+    padding: 55px 110px 54px;
```
andresgnlez: change for `padding: 55px 110px;` nobody is going to die for a pixel :P

andresgnlez:
@hectoruch this PR has conflicts as you can see. Please fix them as soon as possible to avoid more conflicts. Also check some comments I did, ask me if you have doubts!

PR #124, opened 2016-08-12
Improve rendering of canvas and track layers

nerik:
This has many shortcomings (see comments) but I think overall it's an improvement.

screen shot 2016-08-12 at 13 29 11

screen shot 2016-08-12 at 12 58 41


app/src/components/Layers/TrackLayer.jsx
```diff
@@ -150,7 +133,7 @@ const createTrackLayer = function (google) {
    * @param filters
    * @param vesselTrackDisplayMode
    */
-  TrackLayer.prototype.drawTile = function (data, series, filters, vesselTrackDisplayMode) {
+  TrackLayer.prototype.drawTile = function (data, series, filters, vesselTrackDisplayMode, weight, zoom) {
```
nerik: Performance could be optimized here by:
- using canvas.Path2D to batch canvas commands (only supported by FF and Chrome, has a polyfill)
- inline function calls (at the cost of code readabilty, see if we can automate this with the transpiler)

app/src/components/Map.jsx
```diff
@@ -136,13 +136,18 @@ class Map extends Component {
...
     }
     this.state.trackLayer.recalculatePosition();
-
     this.state.trackLayer.drawTile(
       workProps.vesselTrack.seriesGroupData,
       workProps.vesselTrack.selectedSeries,
       workProps.filters,
-      workProps.map.vesselTrackDisplayMode
+      workProps.map.vesselTrackDisplayMode,
+      // I don't understand why this array has way more points than it should
```
nerik: I need the weight data for track points. How can I get this? `workProps.vesselTrack.seriesGroupData.weight` has like 20000 values while `workProps.vesselTrack.seriesGroupData` has 80 for instance.

app/src/components/Map.jsx
```diff
@@ -136,13 +136,18 @@ class Map extends Component {
...
+      // workProps.vesselTrack.seriesGroupData.weight,
+      1,
+      this.state.zoom
     );
+    // this shouldn't be called when no apparent track shows up, which I don't know how to test
+    this.state.overlay.dim();
```
nerik: We want the canvas layer to dim when a user shows a track. Problem is, for some reason `workProps.vesselTrack.seriesGroupData` / `workProps.vesselTrack.selectedSeries` always has something even when user clicked in the middle of the ocean where there are no vessels at all, so the canvas layer will dim whenever she clicks on the map.
Other (UI) thing: can we deselect a track? Right now once you select a track it's dimmed forever.

app/src/util/canvasPointRendering.js
```diff
@@ -0,0 +1,11 @@
+export default {
+  getRadius: (weight, /* sigma */ zoom) => {
+    // seems like sigma is already zoom dependant?
+    // const normalizedSigma = Math.max(1, 3 * (Math.log(1+sigma)/zoomMult));
```
nerik: ATM I used the value encoded in symbol size is weight, while it should be sigma. I have to investigate what this sigma value means and how it works - it seems that its value changes depending on the zoom level. Could that be possible? I pulled my hair while trying to implement zoom dependent styling because it seems so.

app/src/components/Map.jsx
tiagojsag: that's something we need to add: when a user clicks on a non-vessel pixel, hide the track + clear this state. might also be related to the issue where we need to filter the clickable point by innerbounds, instead of the outter bounds

nerik: Okay, looking at this now.

tiagojsag:
Seems that if I open the map and change the inner bounds on the time slider, the refresh fails

nerik:
@tiagojsag yes it does this sometimes but I think it's unrelated. In general I've always had some tiles not refreshing.

PR #128, opened 2016-08-12
[MAP] Feature/search improvement

clementprdhomme:
This PR improves the search panel by:
- making the results scrollable if more than three
- adding a "no result" label
- taking care of not sending a request with an empty search term
- adding a "clear" button
- improving accessibility and adding new styles
- fixing an issue where the advanced filters couldn't be opened at the same time as the seeing the results

The search panel will still need more work:
- some design/animation refinements
- a spinning loader to tell the results are loading
- moving the advanced filter to a level-based layer


app/src/actions/search.js
```diff
@@ -1,30 +1,43 @@
...
+      count: -1
+    }
+  };
+}
+
 export function getSearchResults(searchTerm) {
   return (dispatch, getState) => {
     const state = getState();
 
     if (!state.user.token) {
```
tiagojsag: this is totally fine, but a slightly better way to do it is to return early when requirements fail: https://github.com/Vizzuality/GlobalFishingWatch/blob/master/app/src/actions/vesselInfo.js#L7

by doing so, the actual logic has less "if" "else" and whatever around it, making it easier to understand.

clementprdhomme: I agree with you. I started with a return statement but eslint wouldn't let me because of inconsistent statements. Maybe we can skip the rule.

clementprdhomme:
Probably not useful anymore.

PR #131, opened 2016-08-13
Random cover image

tiagojsag:
Random cover image

This will be useful at least for the blog post pages, were we should avoid having always the same image for all posts. This also allows us to pass a prop to the cover with the image, so it supports easy changing the cover image being used.


app/src/components/Shared/CoverPrimary.jsx
```diff
@@ -3,33 +3,74 @@ import CoverPage from '../../../styles/components/c-cover-page.scss';
...
+          <Header />
+          <h1 className={CoverPage['cover-main-title']}>
+            {this.props.title}
+          </h1>
+          <p className={CoverPage['cover-main-subtitle']}>
+            {this.props.subtitle}
+          </p>
+          <div className={CoverPage['footer-header']}>
+            <div>
+              <img className={CoverPage['ldf-logo']} src={LogoLDF} alt="logo" />
```
andresgnlez: `alt` property kind of pointless. Better being more specific,  ie: `ldf-logo`

andresgnlez:
I think background's method is quite orthodox but it works. We can keep it meanwhile and find another solution later.

PR #132, opened 2016-08-14
Blog pagination

tiagojsag:
Actually working pagination.
Loads page count from API and renders paginator accordingly.
Support URL /blog?page=<int> to load specific page

### DIFF
--- a/app/src/components/Blog.jsx
+++ b/app/src/components/Blog.jsx
@@ -1,15 +1,49 @@
 import React, { Component } from 'react';
 import Footer from './Shared/Footer';
 import { Link } from 'react-router';
-import listposts from '../../styles/components/c-list-posts.scss';
+import listPosts from '../../styles/components/c-list-posts.scss';
 import CoverPrimary from './Shared/CoverPrimary';
 import PaginationBlog from './Blog/Pagination';
 import boxtriangle from '../../assets/icons/box_triangle.svg';
 
 class Blog extends Component {
 
+  constructor(props) {
+    super(props);
+
+    let currentPage = 1;
+    if (props.queryParams && props.queryParams.page) {
+      currentPage = parseInt(props.queryParams.page, 10);
+    }
+
+    this.state = {
+      totalPageCount: 0,
+      currentPage
+    };
+
+    this.onPageChange = this.onPageChange.bind(this);
+  }
+
   componentDidMount() {
-    this.props.getRecentPost();
+    this.props.getRecentPost(this.state.currentPage);
+  }
+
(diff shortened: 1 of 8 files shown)


app/src/components/Blog.jsx
```diff
@@ -1,17 +1,36 @@
...
 
 class Blog extends Component {
 
+  constructor(props) {
+    super(props);
+    this.state = {
+      totalPageCount: 0
+    };
+
+    this.boundOnPageChange = this.onPageChange.bind(this);
```
nerik: Just do:
```
this.onPageChange = this.onPageChange.bind(this);
```

app/src/components/Blog/Pagination.jsx
```diff
@@ -6,17 +6,30 @@ import ReactPaginate from 'react-paginate';
...
-          pageNum={10}
-          activeClassName={pagination['page-selected']}
-        />
-      </div>
-    </section>);
+    if (this.props.pages === 0) {
+      return null;
+    }
+
+    return (
```
nerik: This component doesn't render current page correctly at page load, when page > 1.

nerik: Use `initialSelected`

app/src/containers/Blog.js
```diff
@@ -7,8 +7,11 @@ const mapStateToProps = (state) => ({
   recentPost: state.blog.recentPost
 });
 
-const mapDispatchToProps = (dispatch) => ({
-  getRecentPost: () => dispatch(getRecentPost())
-});
+const mapDispatchToProps = (dispatch, { location }) => {
```
nerik: nice

app/src/components/Blog.jsx
tiagojsag: that's what we've been doing, yes, but to me that's kind of a smell. We are basically having two things with the same name, the bound and the unbound function. Thoughts?

nerik: We are never using the unbound function. How do you think it's a smell?

tiagojsag: maybe it's a im-a-noob-js-dev-so-this-is-not-really-a-smell kind of smell, you know. It's not about how it's used, it's about code readability. The JS interpreter overloads the unbound method with the bound one, just that a human might confuse one for another when reading the code.

nerik: But again, confusing them don't really matter since we only mean to use one. If you go with 2 names, I'd say pageChangeBound instead of boundOnPageChange.

PR #134, opened 2016-08-16
Feature/add git hook helpers

tiagojsag:
Added a git hook manager + some hooks:
- on merge, if your .env file needs to be updated, you get a warning
- on merge, if you package.json changed, npm install is executed automatically
- on commit, js lint runs (cancels the commit if it failed)
- on commit, scss lint runs for commiting files (not blocker)

to enable, pull + run `./bin/git/init-hooks`
more info: https://github.com/tiagojsag/githooks


bin/git/hooks/post-merge-warn-env-changes.sh
```diff
@@ -0,0 +1,24 @@
...
+changed_files="$(git diff-tree -r --name-only --no-commit-id ORIG_HEAD HEAD)"
+
+check_run() {
+	echo "$changed_files" | grep --quiet "$1" && eval "$2"
+}
+
+warn_env_update() {
+  echo ""
+	echo "========================================================"
+	echo "================= ENV FILE UPDATED ====================="
```
nerik: Can we have this blinking? :D

tiagojsag: https://github.com/sjungwirth/githooks/blob/master/bin/git/hooks/pre-commit-reject-debug

original file I copied it from. no blinking, but I can throw in a badass skull if you want ;)

nerik:
Nice. Haven't had the time to run those hooks, but a few comments:
- I'm still not sure about running linters at pre-commit. They were only at pre-push and from what I've seen in the last 2 weeks, ppl are already either skipping hooks or did not add them. I'm afraid this is gonna be too intrusive, maybe we should discuss as a team? Or it's bikeshedding, dunno.
- Don't forget to inform everyon to run the init script :)
Otherwise lgtm.

bin/git/hooks/post-merge-warn-env-changes.sh
nerik: YES

tiagojsag:
Well people can always skip hooks, but this way there's no longer a "oh, I forgot" or other similar excuses. you run a command once, and all future hook changes are loaded automatically from upstream.

The hooks are only blocking js errors, which we already have removed, so if the js linter blocks, someone, it means bad code is creeping in.

The scss linter is only informative for now, it is not blocking anything (yet)

PR #137, opened 2016-08-16
Feature/accordion styles rebase

andresgnlez:


andresgnlez:
I added all of you because there are many base styles we are not using at this moment. We haven't setted wraps, some layout sizes and common styles in `_.base.css`. I have added some of them in this PR. Feel free to continue adding anything you find useful and necessary, but keep in mind this settings before write your own rules. 

There is also some comments in the code, maybe you can help me to choose what to do with these stuff :)

Let's try to keep clean and organized `base` and `settings` files, after all they are the pillars of our CSS structure! Well, and all files in general hahaha.

Thanks!!

app/styles/components/shared/c-content-accordion.scss
```diff
@@ -1,59 +1,43 @@
...
-      padding-bottom: 60px;
+    & + .accordion-item {
+      border-top: 2px solid $color-10;
     }
   }
 
-  .accordion-item .item-title {
+  > .accordion-item .item-title {
     color: $color-19;
     font-size: 19px;
```
clementprdhomme: Could you change this property too? We need to use the variables.

PR #138, opened 2016-08-16
Redirect to login on URL param

tiagojsag:
This is sort of an obscure feature resulting from the following flow:
- When a user registers as a new user on Salesforce, he goes through all the process, which ends up in an email with a link confirming his registration. That link will take the user to the map. Problem is that, at this point, the user is logged in in SalesForce, but our app does not have the oauth token, so if he goes into the map, he will see the login message. So, we add the new flag to the URL in the email and, if the user comes into the map with it and is not logged in, he is redirected to the login page on SalesForce, which automatically redirects him back to us with the token


app/src/containers/Map.js
```diff
@@ -12,20 +12,23 @@ import {
...
 import { updateFilters } from '../actions/filters';
 import { getVesselTrack, setCurrentVessel } from '../actions/vesselInfo';
+import { login } from '../actions/user';
 import { RESET_VESSEL_DETAILS } from '../actions';
 
 const mapStateToProps = (state) => ({
   map: state.map,
   vesselTrack: state.vesselInfo.track,
   filters: state.filters,
-  loggedUser: state.user.loggedUser,
```
tiagojsag: these seem to be leftovers from when the header menu was built into the map.

nerik:
You have routing logic inside components, it's a smell. Even more when duplicated. And adding boilerplate to the 500 lines long Map.jsx makes me nervous. It should be in containers (or even upper, Routes.jsx?)
Problem is, because of react-router black magic, the `location` part of the store is only accessible in 'routed' components.
Unless we have some `react-router` expert around, I suggest using native `document.location.search` with a regex. Not optimal but it would still be much better.

app/src/routes.jsx
```diff
@@ -23,7 +23,7 @@ class Routes extends Component {
       <Route path="/" component={AppContainer}>
         <IndexRoute component={HomeContainer} />
         {!EMBED_MAP_URL && <Route path="map" component={MapContainer} />}
-        {EMBED_MAP_URL && <Route path="map" component={StubMapContainer} />}
+        {EMBED_MAP_URL && <Route path="map" component={MapIFrameContainer} />}
```
nerik: Probably a matter of style, but I'd rather have :
```
const map = (EMBED_MAP_URL) ? MapIFrameContainer : MapContainer;
...
<Route path="map" component={map} />
```

nerik:
Hey, in fact, this: https://github.com/Vizzuality/GlobalFishingWatch/pull/132/files#diff-f2621ea7f3234facadb6685f7fe55c59R10
should actually do the trick :)

nerik:
Okay, after discussion with @j8seangel :
You need to create a MapAuthContainer, that will both handle:
- whether you should instanciate MapContainer or MapIframeContainer (currently in routes.jsx) 
- whether the user should be redirected to the login page

For the 2nd thing, use the solution I highlighted here:
https://github.com/Vizzuality/GlobalFishingWatch/pull/132/files#diff-f2621ea7f3234facadb6685f7fe55c59R10
This will allow using the location part of the store within a container, instead of having this in a component.

nerik:
Let me know if you need help with this.

j8seangel:
Let me know if you need help too :)

nerik:
@j8seangel do you wanna have a look?

j8seangel:
I don't like it too much because you container looks like a component.

For me is better to separate the logic (in the container) and the render (in the component), in that way you don't need the context and looks easier for me so I would make something like this:

- Container

```
import AuthMap from '../components/AuthMap';

import { login } from '../actions/user';

const mapStateToProps = (state, { location }) => ({
  isLogged: !state.user.token && location.query && location.query.redirect_login || false,
  location
});

const mapDispatchToProps = (dispatch) => ({
  login: () => dispatch(login())
});

export default connect(mapStateToProps, mapDispatchToProps)(AuthMap);
```

- Component

```
import MapContainer from '../containers/Map';
import MapIFrame from '../components/MapIFrame';

class AuthMap extends Component {
  componentWillMount() {
    if (!this.props.isLogged) {
      this.props.login();
    }
  }
  render() {
    return (EMBED_MAP_URL) ? <MapIFrame /> : <MapContainer location={this.props.location} />;
  }
}

AuthMap.propTypes = {
  isLogged: React.PropTypes.bool,
  location: React.PropTypes.object,
  login: React.PropTypes.func
};

export default AuthMap;
```

nerik:
Allow me to disagree :)
If you look at your container, it contains *one single line of logic* (the one with isLogged). The rest is boilerplate!
Even though my original component has a render method, it's not a presentational component, as it deals solely with mounting one subcomponent or one another.
Therefore I think my version makes sense and is more terse. I need the context because I'm not using `connect` (it's not needed here), which itself uses context behind the scenes.
Overall I think we need to be flexible and aim at code terseness and expressivity instead of following patterns.

tiagojsag:
sorry guys, I really need this up on staging.

PR #139, opened 2016-08-17
added app favicon

andresgnlez:


clementprdhomme:
👍

Could you run ImageOptim and SVGO on the files so we reduce their size?

andresgnlez:
Good point! Already done.

PR #141, opened 2016-08-17
divisor lines now fits the whole width of the screen

andresgnlez:


app/src/components/Shared/ContentAccordion.jsx
```diff
@@ -51,7 +51,7 @@ class ContentAccordion extends Component {
     return (
       <Accordion
         allowMultiple={false}
-        activeItems={[-1]}
+        activeItems={0}
```
tiagojsag: do we want to have the first open by default? feels a bit weird to me tbh but please check with Dani

PR #150, opened 2016-08-18
Removes the duplicates styles

clementprdhomme:
Previously, all the styles would be duplicated due to multiple imports. This PR solves it by:
* removing the reference to `styles.css` inside the HTML file as it is automatically included by webpack
* removing the import of `base.scss` in `index.scss` as it is already imported in `application.scss`

Note: `index.scss` should be entirely removed as the styles for each component should be imported via webpack/JS but I left them untouched to prevent breaking anything at the moment. I'll add a task to Pivotal.


tiagojsag:
Our map looks pretty epic with this

nerik:
@tiagojsag @clementprdhomme keeping the reference to base.scss left everything intact. I don't think keeping multiple refs to the same scss files don't matter anyway, because the sass compiler will not duplicate styles.

clementprdhomme:
We still have duplicates for the base styles 😕.

nerik:
This last commit:
- removes duplicated global styles (removes global styles loaeded at react component level)
- removes index.scss
- puts all global styles in a file that is loaded once
- removes that ugly div > div > div trick in base.scss

I was careful not to introduce any regressions in the styles (including in the map), but it's worth double checking. @clementprdhomme would you please have a look at this? Thanks :)

hectoruch:
Something is wrong on responsive styles here.

image

nerik:
Thanks @hectoruch, good catch.

clementprdhomme:
I haven't experienced any issue so far. Is it ready to merge?

PR #154, opened 2016-08-19
Fix styles menu errors and add wrap grid

hectoruch:


tiagojsag:
Tested and it seems ok

PR #155, opened 2016-08-19
IFrame map footer

nerik:
screen shot 2016-08-19 at 14 28 39


tiagojsag:
I added a couple of changes of my own:
- the idea on the design for the logo opacity was to "break" a bit, as they were much brighter than the map, which was dark blue. Since we don't have that basemap on the iframe map (google default basemap), I am removing it, as it looked weird.
- same thing for terms and google maps/carto attribution
- added a width:100% so that the footer is full wide on the screen

PR #156, opened 2016-08-22
Update text container styles on cover pages

hectoruch:


app/styles/components/c-cover-page.scss
```diff
@@ -21,7 +25,6 @@
     line-height: .85;
     margin-bottom: 0;
     margin-top: 0;
-    max-width: 1280px;
     padding-top: 170px;
     position: relative;
     top: 0;
```
andresgnlez: Are `position` and `top` properties necessary? I don't think so.

PR #157, opened 2016-08-22
map login back button

nerik:
screen shot 2016-08-22 at 11 02 32

Still very far from designs, lmk if you want me to go there

screen shot 2016-08-22 at 11 04 18


tiagojsag:
tested, looks good to me

andresgnlez:
@nerik we are using global styles for links (check _base.scss). It should be valid for your link adding the class -underline.

PR #159, opened 2016-08-22
added some styles and markup to password forms

andresgnlez:


public/form/css/styles.css
```diff
@@ -403,6 +420,42 @@ section.ty {
...
+  line-height: 2;
+}
+
+@media screen and (min-width: 768px) {
+  .password-hint {
+    text-align: center;
+  }
+}
+
+.password-hint span {
```
nerik: Remove this and use a `p` instead?

public/form/newpassword.html
```diff
@@ -73,18 +73,24 @@ <h1 class="logo">Global Fishing Watch</h1>
...
           <input type="text" id="gfw_password" placeholder="********" required />
         </div>
         <div class="item">
-          <label for="gfw_repeat_password">Repeat Password *</label>
+          <label for="gfw_repeat_password">Verify new password *</label>
           <input type="text" id="gfw_repeat_password" placeholder="********" required />
         </div>
 			</div>
-			<input type="submit" value="Save"/>
+			<input type="submit" value="set password"/>
```
nerik: Should we change this in changepassword.html as well?

PR #160, opened 2016-08-22
added section headings to FAQ

nerik:
screen shot 2016-08-22 at 11 35 39

### DIFF
--- a/app/styles/_base.scss
+++ b/app/styles/_base.scss
@@ -81,3 +81,14 @@ a {
     }
   }
 }
+
+.section-title {
+  color: $color-9;
+  display: block;
+  font-size: $font-size-xx-big;
+  font-weight: $font-weight-normal;
+  line-height: 1.05;
+  margin: 60px 0 18px;
+  max-width: 700px;
+  width: 100%;
+}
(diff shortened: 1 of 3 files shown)


app/styles/_base.scss
```diff
@@ -81,3 +81,14 @@ a {
...
+.section-title {
+  color: $color-9;
+  display: block;
+  font-size: $font-size-xx-big;
+  font-weight: $font-weight-normal;
+  line-height: 1.05;
+  margin: 60px 0 18px;
+  max-width: 700px;
+  width: 100%;
+}
```
nerik: This was almost duplicated from blog's c-list-posts.scss. It should probably be a base style, not at component level.

andresgnlez:
@nerik we are using global styles for links (check  `_base.scss`). It should be valid for your link adding the class `-underline`.

app/styles/_base.scss
andresgnlez: Agree. We should do something with all those common styles...

andresgnlez:
Ignore last comment. Wrong PR.

PR #161, opened 2016-08-22
Name and email filled in by default

nerik:
in the contact form, when logged in, name and email filled in by default


tiagojsag:
There was a React warning when you are logged in on the contact page and then log out. My commit addresses it. otherwise looks good

PR #164, opened 2016-08-22
redesign articles pubs page

andresgnlez:


app/src/components/ArticlesPublications.jsx
```diff
@@ -23,26 +51,40 @@ class ArticlesPublications extends Component {
...
+            </h2>
+            <div className={PubsArticlesLayoutStyles.grid}>
+              {publications}
+            </div>
           </section>
         );
       }
     }
 
     if (articlesPublications && articlesPublications.publications) {
```
tiagojsag: checking for publications, using articles (same thing above)

app/src/components/ArticlesPublications.jsx
```diff
@@ -23,26 +51,40 @@ class ArticlesPublications extends Component {
     let pageContent = [];
 
     if (articlesPublications && articlesPublications.articles) {
-      const articles = articlesPublications.articles.map(this.renderElement);
+      const publications = articlesPublications.publications.map(this.renderElement);
 
-      if (articles && articles.length) {
```
tiagojsag: out of curiosity, why did you change these?

andresgnlez: The name? Yes, it less confuse than having two `publications` variables, in my opinion.

tiagojsag: no, in here we had articles and you replaced with publications. In line 76, we had publications and you replaced with articles. code is the same, you just changed the order for some reason

andresgnlez: Publications come first in the design.

tiagojsag:
Tested it and it looks good

app/src/components/Shared/Rhombus.jsx
```diff
@@ -9,7 +9,7 @@ class Rhombus extends Component {
     const direction = this.props.direction ? RhombusStyles[this.props.direction] : '';
     const color = this.props.color ? RhombusStyles[this.props.color] : '';
 
-    return `${RhombusStyles['c-link-to']} ${direction} ${color}`;
+    return `${RhombusStyles['c-rhombus']} ${direction} ${color}`;
```
nerik: The whole method body could be replaced by:
```
classnames(RhombusStyles['c-rhombus'], RhombusStyles[this.props.direction], RhombusStyles[this.props.color]);
```

app/styles/components/c-publication-article.scss
```diff
@@ -0,0 +1,44 @@
...
+
+  > .title a {
+    color: $color-9;
+    font-family: $font-family-1;
+    font-size: $font-size-x-medium;
+    font-weight: $font-weight-light;
+    line-height: 1.4;
+  }
+
+  > .description p {
```
nerik: Definitely not critical on this page, but avoid styling tags directly, performance is bad.

app/styles/components/c-publication-article.scss
```diff
@@ -0,0 +1,44 @@
...
+    font-family: $font-family-2;
+    font-size: $font-size-small;
+    font-weight: $font-weight-bold;
+    letter-spacing: -.1px;
```
nerik: Why do you need this?

app/styles/layout/l-publications-articles.scss
```diff
@@ -0,0 +1,45 @@
...
+
+  .section-title {
+    margin: 0  0 30px;
+
+    @media #{$mq-tablet} {
+      margin: 0  0 60px;
+    }
+  }
+
+  section {
```
nerik: Re https://github.com/Vizzuality/GlobalFishingWatch/pull/164/files#r75878183

app/styles/layout/l-publications-articles.scss
```diff
@@ -0,0 +1,45 @@
...
+    padding: 55px 0 65px;
+    position: relative;
+
+    &::after {
```
nerik: Why not applying a border-bottom instead?
(you'd need to have `<div className={AppStyles.wrap}>` inside `section`s though)
Not sure about this.

nerik:
Looks good, just a few comments.
I think at some point we'll have to split ArticlesPublications.jsx into subcomponents.

andresgnlez:
Yeah, I agree with you on spliting both kind of publications.

app/styles/layout/l-publications-articles.scss
andresgnlez: I use `::after` because lines have to occupy the whole screen width and you can't do that with `border-bottom`. If you have any suggestion about this, it's welcome :)

I don't use `wrap` inside sections, I only use once and its surrounding them :/

app/styles/components/c-publication-article.scss
andresgnlez: Because those styles are quite different from usual ones. Do you mean this?

nerik: I mean `letter-spacing: -.1px;`, just wondering what's that for.

app/styles/layout/l-publications-articles.scss
nerik: Nevermind then.

PR #166, opened 2016-08-23
Header images

tiagojsag:


tiagojsag:
@andresgnlez can you finish up here please? thanks

andresgnlez:
@tiagojsag done. Are images optimized?

tiagojsag:
yes, but you might want to make them lower res if they are still too big

app/styles/_settings.scss
```diff
@@ -17,6 +17,7 @@ $wrap-desktop: 115px;
 // Font sizes
 $font-size-tiny: 10px;
 $font-size-x-small: 12px;
+$font-size-xs-small: 13px;
```
tiagojsag: honest question: can't we just agree on a standard set of sizes? otherwise this loses total sense

andresgnlez: Give us a couple of days and we'll do a refactor..  😁  Anyway, that size is the UI Kit and spoken with Dani.

PR #167, opened 2016-08-23
setted menu mobile logic

hectoruch:


app/src/components/Shared/MenuMobile.jsx
```diff
@@ -2,34 +2,53 @@ import React, { Component } from 'react';
...
-          <li>FAQ</li>
-          <li>Tutorials</li>
-          <li>Definitions</li>
-          <li>About</li>
-        </ul>
-        <ul className={menuMobile['submenu-mobile']}>
-          <li>The Project</li>
-          <li>Partners</li>
-          <li>Contact Us</li>
+    console.log(this);
```
tiagojsag: I'm gonna start a "you left dev code on a PR" jar, and you have to put money on it every time this happens

app/src/containers/MenuMobile.js
```diff
@@ -0,0 +1,15 @@
+import { connect } from 'react-redux';
+import MenuMobile from '../components/Shared/MenuMobile';
+import { setVisibleMenu } from '../actions/appearence';
+
+const mapStateToProps = (state) => ({
+  menuVisible: state.appearance.menuVisible
```
tiagojsag: this can be component state, I dont think we need this in redux state container

app/src/components/Shared/MenuMobile.jsx
andresgnlez: wow.. doesn't meant test should have stopped him from pushing...?

app/src/containers/MenuMobile.js
hectoruch: @tiagojsag 
But if that is a component state...who can we change that if we call the function on header?
I mean that is neccesary not?

tiagojsag: The MenuMobile knows and needs to know if it's open or closed, so it's always stateful. Difference is is the state is internal (this.state) or external (this.props). Nobody outside of the MenuMobile uses that state. External components only care about opening or closing it. So the interface to the component (mapDispatchToProps) should have an "open" and a "close" mappings that external elements (hamburguer icon) can call. They don't care about the state, neither should they

@clementprdhomme thoughts ?

app/src/components/Shared/MenuMobile.jsx
tiagojsag: it stops you if you use it => README.md

app/src/containers/MenuMobile.js
clementprdhomme: Let's merge it as it is for now (if that's the only blocking point) and improve it later. We need a bit of refactoring as the desktop and mobile menu don't use the same "architecture". We'll be able to remove the redux state after the refactoring.

tiagojsag: @andresgnlez we can leave this out for now, as clement suggested, but lets create a ticket to improve the overall usage of redux

andresgnlez: Ok, I agree.

tiagojsag: ok we need to change this now. I'll push a commit in a sec

PR #168, opened 2016-08-23
remove access token from hash on logout

nerik:


app/src/actions/user.js
```diff
@@ -51,6 +51,7 @@ export function logout() {
     dispatch({
       type: LOGOUT
     });
+    window.location.hash = window.location.hash.replace(/#access_token=([a-zA-Z0-9.\-_]*)/g, '');
```
nerik: I could use the regex in App.jsx, but it looks like it's broken, and I don't dare touching it

tiagojsag: regexes: adding magic and tech debt since 1956 (I googled that)

nerik: I don't get what the `(&[a-z=])?` bit is doing here https://github.com/Vizzuality/GlobalFishingWatch/blob/master/app/src/components/App.jsx#L4
(capturing one single char after & ?)

PR #173, opened 2016-08-23
create styles for tutorial page

hectoruch:


app/src/components/Tutorials.jsx
```diff
@@ -1,6 +1,8 @@
 import React, { Component } from 'react';
 import CoverPrimary from './Shared/CoverPrimary';
 import Footer from './Shared/Footer';
+import baseStyle from '../../styles/_base.scss';
```
andresgnlez: Don't import `_base` directly, use `application` instead, please.

PR #185, opened 2016-08-25
Fix/map height

andresgnlez:


andresgnlez:
@hectoruch just take a look and merge :)

tiagojsag:
what's the problem here? this seems to only affect the map taht we are not using right now...

andresgnlez:
@tiagojsag none. I assigned you unintentionally, I'm sorry! It's just a cleaning  and fix of map styles.

tiagojsag:
Problem is not me, problem is that we should avoid getting sidetracked on tasks that are not vital right now. I am all up for improvements here and there, but if something is not vital for now, we can add to pivotal and fix it later on ;)

andresgnlez:
Yeah, I understand. I was fixing a little bug with the height and I saw the whole file failing `sass-lint`, that's why I changed.

PR #190, opened 2016-08-25
moved step list to homepage

andresgnlez:


tiagojsag:
Do we have/need licenses for those 3 images?

andresgnlez:
Those images are CC0 (creative commons), they don't need credit.

PR #191, opened 2016-08-25
fix footer map on firefox browser

hectoruch:


andresgnlez:
Working on Firefox v48.0.2 (OSX) and Safari 9.1.2.

@hectoruch I haven't checked it out in IE, please check v10, v11 and Edge.

If everything is OK, just merge!

PR #192, opened 2016-08-25
Feature/tool tip text rebase

hectoruch:


andresgnlez:
@hectoruch the "arrow" is showing its "hidden" borders in the inner of the content. We need to fix this.

PR #195, opened 2016-08-25
implemented terms by url

andresgnlez:
- Created `ContentAccordion ` container to allow router management (`push` function).
- When the user clicks on a item, the url will be updated with the new slug entry.
- Back and forward navigation works smoothly.
- Added `slug` property in JSON file: this way we avoid use regexp and other problems generated because inconsistent format. 

The default library didn't include a `componentWillReceiveProps ` function that allows us update the component with new props, that's why we have to create our "own" accordion and overwritting the function to allow us update the component (that allows the accordion being updated if you want to go back to see your las entry).

Probably you will have doubts about this, talk to me if need!

Kudos to Ra for helping me with this!

### DIFF
--- a/app/src/components/Home/CoverPage.jsx
+++ b/app/src/components/Home/CoverPage.jsx
@@ -4,6 +4,7 @@ import CoverPageStyle from '../../../styles/components/c-cover-page.scss';
 import baseStyle from '../../../styles/application.scss';
 import Slider from '../../lib/react-slick.min';
 import Header from '../../containers/Header';
+import { scrollTo } from '../../lib/Utils';
 import BoxTriangleStyle from '../../../styles/components/c-box-triangle.scss';
 import LogoLDF from '../../../assets/logos/ldf_logo.png';
 import sliderBackground1 from '../../../assets/images/background_1.jpg';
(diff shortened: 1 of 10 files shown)


tiagojsag:
We need to scroll down automatically to where the open element is

andresgnlez:
Yeah, you are right. Anyway I think we can merge this (if works fine) and develop the scroll apart. This way we ensure having the terms by url feature available anytime.

I'll write the task down on Pivotal.

tiagojsag:
No, implement it in this branch. The current implementation adds nothing to the user. He might see the definition, he might not. The user story is "the user should see the definition", not "the user should see the definition if it's one of the first or his screen is big enough"

tiagojsag:
Seems like the new accordion class is not needed at all - tested like this and it seems to work.
Please add the scroll feature ASAP so we can merge this.

andresgnlez:
With your commit, Accordion won't update if you go back or forward in hsitory navigation. Teh extended function is needed for this functionality.

tiagojsag:
My commit? which commit? I don't see any commit :P
Hadn't tested that scenario. In any case, add the scroll thing please

andresgnlez:
I think the branch's history is quite mess... (it uploaded first commit again..) but I don't see anything strange. Check it out yourself, please.

app/src/components/Home/CoverPage.jsx
```diff
@@ -66,7 +67,7 @@ class CoverPage extends Component {
     const sliderAttributions = [
       '© OCEANA / Juan Cuetos',
       null,
-      '© Steve De Neef ',
+      '© OCEANA / Eduardo Sorensen',
```
tiagojsag: can you change this back,please? the correct value here is the "steve" one

tiagojsag:
once you change the attribution thing, you can insta-merge, no need for another CR

PR #197, opened 2016-08-26
New background images on login and register page and update logo

hectoruch:
Change:

- Registration background page
- Login background page
- Header logo


public/form/css/styles.css
```diff
@@ -705,23 +716,11 @@ hr {
 }
 
 .bg.log {
-  background-image: url('../img/sky.png');
```
tiagojsag: @clementprdhomme thoughts?

public/form/css/styles.css
```diff
@@ -616,6 +616,17 @@ hr {
   width: 100%;
 }
 
+.attribution-picture{
+  text-align:right;
+}
+
+.attribution-picture span{
+    color: #fff;
+    font-family: "Roboto Slab", serif;
```
tiagojsag: this is not being picked up - my browser shows me the attribution in times new roman

public/form/css/styles.css
```diff
@@ -117,7 +117,7 @@ header {
   max-width: 175px;
   margin: 0;
   text-indent: -10000px;
-  background: url("../img/logo-mobile@2x.png") no-repeat center center;
+  background: url("../img/gfw_logo_hor.svg") no-repeat center center;
```
tiagojsag: please make sure the logo is the same size as in the main site - right now is smaller

public/form/resetpassword.html
```diff
@@ -1,5 +1,6 @@
 <html lang="en">
 <head>
+	<meta charset="utf-8">
```
tiagojsag: in the future, let's avoid changes that are not part of the current task. this may seem harmless, but this is how magical regressions get introduced

andresgnlez: I agree but this time codification was introduced because copyright symbol wasn't displaying properly. So I guess it's related with the task.

tiagojsag: ah ok, my bad

public/form/css/styles.css
hectoruch: I talk with Daniel about that change and he say me that is no necessary have two version of the same image in that case. @tiagojsag @clementprdhomme

PR #201, opened 2016-08-26
Feature/update images sf rebase and styles

hectoruch:
I update styles for header.
That is not like the our app styles because we need change the DOM and also others styles.
But is the best I can do with the DOM exist right now on public pages.
@nerik @clementprdhomme @andresgnlez @tiagojsag


andresgnlez:
ready to merge for me.

tiagojsag:
All pages need to use the two new images - please remove all old background images and make sure all pages have one of the two new.

andresgnlez:
Please, use references "the two new images" doesn't help at all :/ No idea what you are talking about.

public/form/css/styles.css
```diff
@@ -705,23 +747,11 @@ hr {
...
 
 .bg.log {
-  background-image: url('../img/sky.png');
-}
-
-@media only screen and (min-width: 768px) {
-  .bg.log {
-    background-image: url('../img/sky@2x.png');
-  }
+  background-image: url('../img/login.jpg');
```
tiagojsag: @hectoruch please confirm with @clementprdhomme that this is the way to go

tiagojsag:
See that all comments from https://github.com/Vizzuality/GlobalFishingWatch/pull/197 are addressed

tiagojsag:
Please make sure ALL pages have attribution and that the attribution matches the image

andresgnlez:
is this OK for merge, @tiagojsag ?

tiagojsag:
menu seems to be broken on Iphone 6s plus landscape - sometimes it shows up, sometimes it doesnt, sometimes has background, sometimes not... can you confirm it's not just on my phone?

PR #204, opened 2016-08-26
Feature/error pages

andresgnlez:


tiagojsag:
I dont understand why we have the images duplicated. can you explain, please?

tiagojsag:
Missing attribution

tiagojsag:
- [ ] Missing attributions
- [ ] Header style is different (maybe ask dani if we can live without the header menu on both pages, keeping just the logo?)
- [ ] Logo has different sizes

app/src/components/Shared/NoFoundPage.jsx
```diff
@@ -0,0 +1,21 @@
+import React, { Component } from 'react';
+import Header from '../../containers/Header';
+import StatusPage from './StatusPage';
+
+class NotFoundPage extends Component {
```
clementprdhomme: The file is called `NoFoundPage` and and the component `NotFoundPage`. 😉

clementprdhomme:
I don't think we need them to be duplicated. In the Javascript folder, we don't need the image for the page 500, and in `public` we don't need the one for 404.

tiagojsag:
thought so too, but I didnt want to assume too much, being the front noob around here ;)

tiagojsag:
Attribution on the 500 page has incorrect font

tiagojsag:
- [ ] On 500, the mobile header and mobile menu are different than on the rest of the pages
- [ ] The 500 page has login selected
- [ ] On mobile menu, make the 'p' in "Research program" upper case

PR #205, opened 2016-08-26
Fix definitions accordion bug

tiagojsag:


andresgnlez:
Nice catch ;) Merging...!

PR #207, opened 2016-08-29
partners contact us

nerik:
screen shot 2016-08-29 at 11 06 10


app/src/components/Partners.jsx
```diff
@@ -23,11 +23,12 @@ class Partners extends Component {
         attribution="US Coast Guard"
         showLDFLogo
       />
-      <div className={StaticPageStyles['l-static-page']}>
```
andresgnlez: Why did you remove the layout?

nerik: The margin-bottom added a white stripe because the last section is now blue, which looked rid. So I needed to remove the margin-bottom, so eventually moving the margin-top to the first section instead of the whole page made more sense.

PR #211, opened 2016-08-29
Update link styles on faq page

hectoruch:
Change faq.json with the underline link class.

### DIFF
--- a/app/styles/_base.scss
+++ b/app/styles/_base.scss
@@ -98,7 +98,27 @@ a {
       top: auto;
     }
   }
+
+  :global &.-underline {
+    color: $color-12;
+    font-weight: $font-weight-bold;
+    position: relative;
+
+    &::after {
+      background-color: $color-6;
+      bottom: 0;
+      content: '';
+      display: inline-block;
+      height: 1px;
+      left: 0;
+      position: absolute;
+      right: 0;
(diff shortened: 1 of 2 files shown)


app/styles/_base.scss
```diff
@@ -98,8 +98,28 @@ a {
       top: auto;
     }
   }
+
+  :global &.-underline {
```
tiagojsag: https://github.com/Vizzuality/GlobalFishingWatch/pull/211/files#diff-da132c0d50c9cd8406b6391530936e2bR84

cant we DRY this?

andresgnlez: @hectoruch did you make it global because of `faq.json`?

hectoruch: I just :global because on json file can not add "className". That way we can use -underline class for two ways.

hectoruch:
I think that is ready to merge @tiagojsag

app/styles/_base.scss
tiagojsag: @andresgnlez there's no way here to NOT have the same 20 lines duplicated just to have the "global"?

hectoruch: @tiagojsag  I was talking with Raul. And is the only way to do that.

PR #212, opened 2016-08-29
fixed contact us form validation

nerik:
- invalid form could be submitted in Safari: fixed
- added a red border on invalid form fields
- form could be submitted when question type was not selected: fixed


app/src/components/ContactUs/ContactUsForm.jsx
```diff
@@ -16,6 +16,7 @@ class ContactUsForm extends Component {
 
     this.handleFormSubmit = this.handleFormSubmit.bind(this);
     this.handleChange = this.handleChange.bind(this);
+    this.onSubmitClicked = this.onSubmitClicked.bind(this);
```
andresgnlez: avoid binding functions and use arrow functions instead, please.

PR #218, opened 2016-08-30
responsive styles for FAQ and Definitions pages

andresgnlez:


tiagojsag:
- [ ] title on FAQ page is cut on iphone 6 plus portrait
- [ ] title on Definition page is cut on iphone 6 plus portrait
- [ ] accordion acting funny on iOS
- [ ] autoscroll on definitions page sometimes results in weird states where the info clicked is not visible
- [ ] FAQ section titles could use a bit more spacing on top

andresgnlez:
Work continued at #247

PR #220, opened 2016-08-30
fix style errors and add new images

hectoruch:
Fix:

- Cover page subtitle styles
- Update new svg logos
- Change cursor on different options at header
- Change font style on footer and links


andresgnlez:
Pivotal task: https://www.pivotaltracker.com/story/show/129439905

andresgnlez:
Rebase on #225

PR #222, opened 2016-08-31
Extract home slider content to json

tiagojsag:
Update your .env before testing


app/src/components/Home/CoverPage.jsx
```diff
@@ -40,106 +42,98 @@ class CoverPage extends Component {
...
+    let coverEntriesContent = (
+      <div>
+        <div key={-1}>
+          <div className={CoverPageStyle['contain-text-cover']}>
+            <h1 className={CoverPageStyle.stub}>&nbsp;<br />&nbsp;</h1>
+            <p>&nbsp;<br />&nbsp;</p>
+          </div>
+        </div>
+      </div>
+    );
```
andresgnlez: Is this the default output if there aren't slides to show...? I don't like it so much..

app/src/containers/CoverPage.js
```diff
@@ -0,0 +1,13 @@
+import { connect } from 'react-redux';
+import CoverPage from '../components/Home/CoverPage';
+import { getCoverPageEntries } from '../actions/coverPage';
+
+const mapStateToProps = (state) => ({
+  coverPageEntries: state.coverPageEntries
+});
```
andresgnlez: Is the state need it? I don't think we save the status of the coverpage at any moment.

app/styles/components/c-cover-page.scss
```diff
@@ -51,6 +51,10 @@
     }
   }
 
+  .stub::after{
+    height: 0px;
+  }
+
```
andresgnlez: Explain please.

app/src/components/Home/CoverPage.jsx
tiagojsag: The idea here is that, if the json is hosted somewhere else, it might take a bit to load. So I tried to make something emtpy-ish but that, at the same time, doesn't use a loading indicator or something that causes a big change when we load the actual content of the slider

app/styles/components/c-cover-page.scss
tiagojsag: when loading the default "empty" slide, the empty <h1> would still have the red :after element. this removes that just for that stub slide

app/src/containers/CoverPage.js
tiagojsag: we have a reducer for this, which is used by the coverPage action, which is used by the coverPage component on load

app/src/components/Home/CoverPage.jsx
andresgnlez: I think it should be easier to implement the loader and make it more consistent with the rest of the site. What do you think, @clementprdhomme ?

app/src/reducers/coverPage.js
```diff
@@ -0,0 +1,11 @@
+const initialState = null;
+import { GET_COVER_PAGE_ENTRIES } from '../actions';
+
+export default function (state = initialState, action) {
+  switch (action.type) {
+    case GET_COVER_PAGE_ENTRIES:
+      return action.payload;
```
clementprdhomme: A reducer **must** always return a copy of the original state with modifications. Example:
```javascript
switch (action.type) {
  case VESSEL_INIT:
    return Object.assign({}, state, action.payload);
    break;
}
```

tiagojsag: The actual requirement is that a reducer should never return a modified version of the previous state object, but should always return a fully new state object. That's why we use Object.assign, to create a copy of the previous state. we are achieving the same here.

app/src/components/Home/CoverPage.jsx
clementprdhomme: Dani says we don't need the loader, but we should get rid of this:
```html
<h1 className={CoverPageStyle.stub}>&nbsp;<br />&nbsp;</h1>
<p>&nbsp;<br />&nbsp;</p>
```

and use a special class to maintain the appearance.

app/src/containers/CoverPage.js
clementprdhomme: @andresgnlez It's the way to retrieve Redux's state.

app/src/reducers/coverPage.js
clementprdhomme: You don't use `Object.assign` here. You just delete the previous state.

clementprdhomme:
The feature doesn't work: the slider never loads. Can you check some of the comments. We can help you with that too.

app/src/components/Home/CoverPage.jsx
```diff
@@ -40,106 +42,98 @@ class CoverPage extends Component {
...
+    const sliderAttribution = (sliderBackgrounds.length ?
+      sliderAttributions[this.state.currentSlider] : '© OCEANA / Juan Cuetos');
 
     return (
       <div className={CoverPageStyle['c-cover-page']} style={{ backgroundImage: `url(${sliderBackground})` }}>
         <CoverPagePreloader images={sliderBackgrounds} />
         <div className={CoverPageStyle['layer-cover']}>
           <Header />
           <div className={baseStyle.wrap}>
-            <Slider {...settings}>
```
clementprdhomme: You can't delete this line otherwise slick doesn't work anymore.

app/src/reducers/coverPage.js
tiagojsag: the only previous state here is either null or the previous list of coverpage elements, which will be replaced anyway. If we add more state here, we need to use object.assign, ofc. but not right now. besides, we are using the same thing on faq, definitions, and mybe some other places too

app/src/components/Home/CoverPage.jsx
tiagojsag: I moved it to line 123

clementprdhomme: Sorry 😁🔫.

app/src/reducers/coverPage.js
clementprdhomme: Ok. You're right on that. But we need to make sure that in case of any addition to this file, we need to change this line.

Thanks @nerik for the help.

andresgnlez:
image

Finish it, @clementprdhomme

PR #223, opened 2016-08-31
mobile zoom + android home screen

nerik:


nerik:
@tiagojsag this can be merged anyway, as it disables the zoom on mobile too.

nerik:
For future reference
https://developer.apple.com/library/ios/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html

tiagojsag:
Flagging this as WIP until we have feedback from the customer

nerik:
Just completing what @andresgnlez did, this is now ready for home screen on Android.
User-zoom is disabled on the whole site, but this doesn't affect the map iframe. We'll probably have to revisit this when using our own map view.
Also: on the map footer, that mention: `Map data ©2016 Google, INEGI Imagery ©2016 NASA, TerraMetrics` is too long on xs screens and breaks the layout + causes a vertical scroll on the page - @tiagojsag could you please check with the client to see if we can shorten this mention or make it available as a tooltip on phone screens?

PR #224, opened 2016-08-31
Analytics

nerik:


nerik:
@tiagojsag please make sure that the client is aware that this is only loaded once (SPA). If they want per-page analytics this is gonna need some more work from us.

tiagojsag:
then we need to do that work :S

nerik:
@tiagojsag GA is implemented. Please ask the client for their tracking code and add it to GA_TRACKING_CODE env.

nerik:
Also, this can be merged as is.

tiagojsag:
Looks good - only need documentation for the new env var on the readme, then you can merge

PR #226, opened 2016-08-31
research program page design

andresgnlez:


app/styles/layout/l-static-page.scss
```diff
@@ -2,4 +2,69 @@
 
 .l-static-page {
   padding: 85px 0 90px;
+
+  .section-page {
+
+    &::after {
```
nerik: This is a duplication of https://github.com/Vizzuality/GlobalFishingWatch/blob/master/app/styles/layout/l-publications-articles.scss#L21
Please abstract into an util class.

clementprdhomme:
Some comments regarding the design, not the code:
* the content should be 700px wide and centered
* the space aren't globally respected. Here are some examples: the space before the first paragraph, the space between each logo and the previous paragraph, the video should be 700px wide to match the content's width, the space between each video and the previous link.


In addition, I noticed you haven't talked to the designer about the links which you should have done. The issue of the red underline has been moved to another task on Pivotal (#129536473) so don't try to fix it in this PR.

andresgnlez:
hey @nerik ! Could you check out the last commit? I unified both section styles in `base`, but not sure if this is the way we have to follow.. Could you take a look and give me your opinion? Thanks mate!

@clementprdhomme styles updated after CR, Thanks!

nerik:
Looks perfectly good to me, @clementprdhomme do you have an opinion on this? Are we using css modules as we should? Should we actually use css modules? Is there life after death?

andresgnlez:
If PR gets green light from @clementprdhomme is ready to merge.

PR #227, opened 2016-08-31
Add logic for load image when user is not login and not load iframe

hectoruch:


app/src/components/MapIFrame.jsx
```diff
@@ -3,8 +3,17 @@ import Modal from './Shared/Modal';
 import NoLogin from '../containers/Map/NoLogin';
 import Header from '../containers/Header';
 import FooterMini from './Shared/FooterMini';
+import mapStyles from '../../styles/components/c-map.scss';
 
 class MapIFrame extends Component {
+
+  constructor(props) {
```
tiagojsag: this does not need to be state, it doesn't change during the component lifecycle. you can use just props.

PR #228, opened 2016-08-31
Add logic to input url on support modal

hectoruch:
Task: https://www.pivotaltracker.com/story/show/129394005


tiagojsag:
This PR is not doing what it should, I wil fix it later myself

PR #229, opened 2016-08-31
better tooltips for mobile

nerik:
https://www.pivotaltracker.com/story/show/129452389


andresgnlez:
image

I think the arrow is moved 1px more and that creates a space between the box and the arrow itself.

andresgnlez:
at least for me, all tooltips show up totally unalienable :/

image

PR #230, opened 2016-08-31
fixes home slider content/background sync

nerik:
The beforeChange issue was actually fixed in a recent version of react-slick.
For some reason we were using a copy of the bundled version of this lib.


nerik:
Also, a take away from this: let's use `react-[whathever-shitty-unmaintained-lib]` only if really necessary. I don't think we needed a lib to show images when clicking on dots. Even more considering that it has 21 open PRs and 85 open issues. @andresgnlez @hectoruch @clementprdhomme @tiagojsag

clementprdhomme:
The reason we used a copy of it is because they fixed another issue in `master` but didn't fixed the npm module. We thus made a copy of the most recent version – at that time.

Regarding you second comment, we will avoid third-party libraries as much as we can in the future. We're dealing with too much issues with them.

nerik:
In the package.json file, you can use any branch of a gh repo, you can even point to a specific commit hash. So normally you should never have to copy a bundled version of a lib. Especially because the package.json way makes this decision explicit, while it's hard or impossible to tell what version of a lib you are using from a lib bundle.

> we will avoid third-party libraries 
IMHO it's really more about "silo" libraries (react-something, gulp-something, etc). They tend to be abandoned very quickly, and for sure they will be when the "main" lib goes out of fashion.
So:
1. ask yourself if you really need a lib
2. if so, look for a vanilla JS (front) / bash (dev pipeline) offering
3. then if nothing's satisfying, use a silo library only if it seems well maintained (more than 1 person pushes commits regularly)

PR #237, opened 2016-09-01
fix accordions issues

nerik:
screen shot 2016-09-01 at 11 55 48


tiagojsag:
- [ ] The animation is a bit "bumpy".
- [ ] The arrow does not turn to the original position when closing the element

nerik:
I fixed above issues by completely removing react-sanfona.
I left the dependencies and the css on index.html because our map will still use those - eventually they should be removed too (I´ll create a pivotal task)

tiagojsag:
on FAQ, when you open the first question, it auto scrolls (do we want this) and it scrolls to the beginning of the answer, not the question.

tiagojsag:
seems like clicking on an open question to close it is not working

nerik:
Should be all good now.

clementprdhomme:
I have two small requests:
* On the FAQ, when clicking on an item from one section to another, I have to click twice in order to open it. Could you change that?
* Could you just add one-line comments regarding the props of the components?

Once the conflicts are fixed and if Tiago is happy about the feature, let's merge it 😄.

PR #240, opened 2016-09-01
fixed tooltips

nerik:
Tooltips are now displayed on both hover and click, so they are accessible on mobile.
The actual link to the definition page is now inside the tooltip.

screen shot 2016-09-01 at 13 51 47

@danielcasius


tiagojsag:
looks good

PR #243, opened 2016-09-01
fix responsive and desktop footer styles

hectoruch:


tiagojsag:
looks good ;)
A small thing: we need to review the breakpoint for the logos. At 1030px +- the 4 logos are all together, which might cause problems

clementprdhomme:
Seems ok. What do you think @tiagojsag?

clementprdhomme:
It's related to this task: https://www.pivotaltracker.com/story/show/129622325

tiagojsag:
Under 770px width, is it possible ot move the LDF logo more to the right?

hectoruch:
@tiagojsag I just change it

tiagojsag:
Functionally it looks perfect now. @hectoruch please clean up that commented code, otherwise can be merged if @clementprdhomme says so

PR #245, opened 2016-09-01
fix hamburger menu on map

nerik:
screen shot 2016-09-01 at 18 43 00
screen shot 2016-09-01 at 18 47 47


nerik:
Note: this relies on the zoom controls being pushed to the bottom, which seems to happens a bit randomly

PR #246, opened 2016-09-01
added blooberg logo and fixed funcding partners styles

nerik:
screen shot 2016-09-01 at 19 23 15
screen shot 2016-09-01 at 19 23 22


tiagojsag:
Let me see if I can ask for a transparent background - that one looks quite crappy

nerik:
And vector, if possible :)

tiagojsag:
@nerik I pushed a commit here with the new, transparent background logo + removed the new padding class you added that I think is no longer needed. Feel free to add it back if it is needed, otherwise rebase and merge

PR #248, opened 2016-09-01
Homepage map image

tiagojsag:


app/styles/layout/l-homepage.scss
```diff
@@ -15,7 +15,7 @@
   background-color: $color-4;
   font-family: $font-family-1;
   padding: 85px 110px 96px;
-
+s
```
andresgnlez: beeee

tiagojsag: Just making sure you were paying attention

PR #249, opened 2016-09-01
Refactoring styles settings

andresgnlez:
Little, but necessary, refactor of our styles settings. Feel free to contribute.


clementprdhomme:
Could you add comments close to the variable to know what they aim to?

I don't like the gap in the colours name going from 13 to 19 but I understand we can't refactor all of that without breaking opened PRs.

clementprdhomme:
This is probably not useful anymore.

PR #247, opened 2016-09-01
Unifying static pages styles

andresgnlez:
Sorry for the amount of files involved in the PR, as you will understand, unify them means modify all of them :/

First commit appears because it's a branch created from `feature/responsive-faq-definitions`, ignore it please: second commit overrides most of this changes.

This PR is based on #218

PR #251, opened 2016-09-02
Static Pages Responsive

andresgnlez:
This PR is based on #247 work so please merge it before merge this.

Pivotal task: https://www.pivotaltracker.com/story/show/129042213

- [x] FAQ page
- [x] Definitions page
- [x] Articles & Publications page
- [x] The Project page
- [x] Tutorials page
- [x] Research Program page
- [x] Contact us page
- [x] Partners page
- [x] Terms of use
- [x] Privacy policy
- [x] Orbcomm


nerik:
Merge before https://github.com/Vizzuality/GlobalFishingWatch/pull/237, or expect a not trivial rebase.

hectoruch:
That PR is ready to review and merge if is possible.
@clementprdhomme @andresgnlez @nerik

app/src/actions/user.js
```diff
@@ -57,7 +57,7 @@ export function logout() {
       type: LOGOUT
     });
     window.location.hash = window.location.hash.replace(/#access_token=([a-zA-Z0-9.\-_]*)/g, '');
-    if (window.location.pathname.match('^\/map')) {
+    if (window.location.pathname.match('/map')) {
```
nerik: Why did you change this?

app/src/components/FAQ.jsx
```diff
@@ -6,6 +6,7 @@ import Accordion from './Shared/Accordion';
 import CoverPrimary from './Shared/CoverPrimary';
 import AppStyles from '../../styles/application.scss';
 import StaticPageStyles from '../../styles/layout/l-static-page.scss';
+import '../../styles/components/shared/c-item-list.scss';
```
nerik: Is this import necessary?

app/src/actions/user.js
hectoruch: Because, eslint warning that use '^\' is not neccesary.

nerik: I don't have that warning. Could you please run `./node_modules/.bin/eslint -c .eslintrc app/src/actions/user.js` and tell me what gives?

nerik: I'm really curious because that regex actually needs the ^

tiagojsag: @hectoruch rule of thumb for life: changing something without understanding the WHY you need to change it is the root of all evils and bugs.

clementprdhomme:
Apart from @nerik's comment, could you remove the `&nbsp;` and replace them with `{' '}` instead in `ResearchProgram.jsx`? The reason is that it doesn't break like a standard whitespace (example: [this line](https://github.com/Vizzuality/GlobalFishingWatch/blob/develop/app/src/components/ResearchProgram.jsx#L261) on mobile).

nerik:
Nice catch, I did the same mistake here and there.

app/src/actions/user.js
hectoruch: Sorry for that, I change right now.

nerik: @hectoruch what was the error?

hectoruch: image
@nerik

hectoruch: image
@nerik

nerik: Aaaaah okayy. Thanks.
This should be:
```
if (window.location.pathname.match('^/map')) {
```

nerik: The ^ caret is necessary, because it means: look for /map at the beginning of the path. The slash didn't need to be escaped.
I don't know why this error was not caught by my eslint.

clementprdhomme: Hum... searching for why eslint sees the line as wrong, I've found this issue: https://github.com/eslint/eslint/issues/6148

It seems that we may need to escape the slash itself: `window.location.pathname.match('^\\/map')`

Do you know more about that @nerik?

clementprdhomme: Oops, didn't see you previous comment. Ok, let's do what @nerik told.

hectoruch: Ok, I will fix it now!

tiagojsag: A few notes:
- in regexes forward slashes need to be escaped (although some regex engines might work in different ways)
- '\\/map' should translate to  '\/map', so not what we are looking for here.
- if you have different results on lint validation, please check your node versions

tiagojsag:
- [ ] FAQ - Line under "Click on the FAQ below to see the answer." is not in the design

hectoruch:
@tiagojsag Last comment I was talk with Dani about that and is fine.

clementprdhomme:
@tiagojsag is right. In `develop`, we fixed the position of the arrow so it is centered with the title and it doesn't overlap the text. Also, its direction is different.

@hectoruch Can you confirm all of that is wanted by @danielcasius? I think we just got a regression here.

clementprdhomme:
Sorry guys for the last comment, I didn't pull the rebase. Is there still any issue on this PR?

clementprdhomme:
Ok, I have some new comments which I hope to be the last before merging the PR:
- [ ] The margin before the grey titles of the FAQ page should be bigger. I think it's just a matter of 5px or so.
- [ ] For the Research Program page, we should take care that the ratio of the videos is not weird on mobiles.
- [ ] On the Contact us page, on iOS, there's an inner shadow in the input fields.
- [ ] On the Partners page, there's twice the title "Founding partners" for two different sections. I'm not sure why. Also, one is written in American English, and the other one in British English.

@hectoruch Could you take care of making these changes so we can go forward?

PR #254, opened 2016-09-02
Fix link on modal window

hectoruch:


tiagojsag:
Compare and fix fonts on:
- Visit home
- header
- text
- link in the text

make sure they all match (right now they dont)

PR #258, opened 2016-09-05
Improves the responsiveness of the accordion and the position of its rhombus

clementprdhomme:
This PR brings small changes to the accordion following Héctor's comments:
* Change the direction of the rhombus to follow the design
* Add a padding to the right of the titles so the rhombus doesn't overlap them
* Vertically center the rhombus

[Pivotal task](https://www.pivotaltracker.com/story/show/129783887)


nerik:
👍

PR #263, opened 2016-09-06
Responsive implementation of Articles and Publications

hectoruch:
Task => https://www.pivotaltracker.com/story/show/129552759


clementprdhomme:
Two remaining issues:
* According to the design, the titles of the articles shouldn't be uppercase.
* On a mobile device, the title "Articles" should be separated from the first article.

hectoruch:
- Fix the distance between title and articles on mobile version
- The titles are Uppercase on JSON file. We need to change it.

PR #264, opened 2016-09-06
Fix message contact logic

hectoruch:
Task => https://www.pivotaltracker.com/story/show/129043553

### DIFF
--- a/app/src/components/ContactUs/ContactUsForm.jsx
+++ b/app/src/components/ContactUs/ContactUsForm.jsx
@@ -8,7 +8,6 @@ class ContactUsForm extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      submitted: false,
       showFormResponse: false,
       classSelect: '',
       disabledOption: false,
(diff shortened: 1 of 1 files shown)


clementprdhomme:
Some comments:
* Please use a button instead of an input for the "Send" button. Use the markup: `<button type="submit">Send</button>`.
* You don't need to have a listener on the button when clicked, because the form will be submitted so you can just listen to the "submit" event on the `form` tag.
* Please don't forget to remove the `submitted` property from the state after you removed the click listener on the button.
* I don't understand why you add a special class on the form when the button is clicked and not when the form is submitted (by entering "enter" for example). It's inconsistent. Also, I don't see the need for the class being added dynamically. Why not add it from the beginning?
* Can you see with Dani the size and the position of the title and intro on the page? It seems really big on mobile.

hectoruch:
About the last comment:
The responsive styles are on that PR => https://github.com/Vizzuality/GlobalFishingWatch/pull/251

The other comments, I will change/fix now.

hectoruch:
Can review @clementprdhomme

clementprdhomme:
One of my previous comment was partially wrong @hectoruch. I told you we don't needed the `validated` to be set dynamically on the form, but if you add it from the beginning, all the fields would initially appear as invalid.

My last PR is about adding the class when the form is submitted. Can you give it a check? If good, we're ready to merge.

PR #265, opened 2016-09-06
Improve accordion logic on mobile

hectoruch:
Task => https://www.pivotaltracker.com/story/show/129786629
Task (2) => https://www.pivotaltracker.com/story/show/129829309 (when you open now an accordion item, title is on top)

### DIFF
--- a/app/src/components/Shared/Accordion.jsx
+++ b/app/src/components/Shared/Accordion.jsx
@@ -11,6 +11,7 @@ class Accordion extends Component {
     this.scrolled = true;
   }
 
+
   onItemClick(index, slug, id) {
     if (window.innerWidth <= 768) {
       this.props.onAccordionItemClick(index, slug, this.props.accordionId);
@@ -20,7 +21,9 @@ class Accordion extends Component {
           return;
         }
         scrollTo(el);
-      }, 10);
+      }, 1100);
+    } else {
+      this.props.onAccordionItemClick(index, slug, this.props.accordionId);
     }
   }
 
--- a/app/src/components/Shared/Accordion.jsx
+++ b/app/src/components/Shared/Accordion.jsx
@@ -3,6 +3,7 @@ import ToolTipJSON from './ToolTipJSON';
 import Rhombus from './Rhombus';
 import Loader from './Loader';
 import $ from 'jquery';
+import classnames from 'classnames';
 import AccordionStyles from '../../../styles/components/shared/c-content-accordion.scss';
 
 class Accordion extends Component {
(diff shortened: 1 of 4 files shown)


app/src/components/Shared/Accordion.jsx
```diff
@@ -4,21 +4,24 @@ import Rhombus from './Rhombus';
 import Loader from './Loader';
 import AccordionStyles from '../../../styles/components/shared/c-content-accordion.scss';
 import { scrollTo } from '../../lib/Utils';
-import $ from 'jquery';
 
 class Accordion extends Component {
   componentDidUpdate() {
```
clementprdhomme: This whole method isn't needed anymore with your new code.

app/src/components/Shared/Accordion.jsx
```diff
@@ -4,21 +4,24 @@ import Rhombus from './Rhombus';
...
     this.scrolled = true;
   }
 
-  onItemClick(index, slug) {
-    this.props.onAccordionItemClick(index, slug, this.props.accordionId);
+  onItemClick(index, slug, id) {
+    if (window.innerWidth <= 768) {
+      this.props.onAccordionItemClick(index, slug, this.props.accordionId);
+      setTimeout(() => {
+        const el = document.getElementById(id);
```
clementprdhomme: This is useless. You can pass the onClick callback the element via `e.currentTarget` so you don't have to search for it. Don't forget to remove the "title" param you added.

clementprdhomme:
I'm still no happy with the solution.

Let's give you a scenario where the user can still get lost:

1. Open the FAQ on your mobile or with the responsive tools of Chrome for example.
2. Open the second item "How can I vary the amount of time I'm looking at?".
3. Scroll to half of its content. Make sure the title disappears from the viewport.
4. Click on the third item (the one right below).

What the user sees is that the third items scrolls down from the top of the screen even if it was located below the second. What happens is that the second item collapses so the third appears at the top of the screen, partially hidden, so that's why it scrolls from the top.

One possible solution is that instead of scrolling to the top of the clicked item, we could maintain it visually at the same place. Let me explain:

1. You open the second item on the FAQ page, and scroll at the middle of its content, as before.
2. You click on the third one, and just before the second collapses, you retrieve the position of the third on screen: `const initialPosition = thirdItem.getBoundingClientRect().top;`
3. Then, you let the second item collapse and retrieve again the position of the third element: `const finalPosition = thirdItem.getBoundingClientRect().top;`.
4. You can now compute the vertical difference between the two: `const offset = finalPosition - initialPosition;`.
5. You scroll the page by the offset so visually the third item hasn't moved: `window.scrollY += offset;`
6. You expand the third item.
7. Eventually, you scroll to the third element so it appears at the top of the page.

The disadvantage of this solution, apart from more logic, is that when the second item will collapse, all of what's before the third item will go down on the page so the third item doesn't move.

Before any change, I want you to talk with @danielcasius about what he thinks of the issue and the solution I propose.

hectoruch:
Daniel says that like it. 
But test it.

clementprdhomme:
Hey Héctor,

I still experience several issues with the accordion:
* I have the same effect where the click item goes up and off the screen before reappearing. See what appears to the red item:
The red items goes off the screen for a bit
* If the clicked item is the first one, then the screen just jumps into its new position without transition (or too quickly):
The animation seems nearly inexistant


Additionally, could you fix these stuffs?
* The code within `componentDidUpdate` seems useless.
* We have a regression with the styles on mobile. The content of the accordion is glued to the title now:
*Before:*
Before, there's a space between them.
*Now:*
Now, it disappeared.

clementprdhomme:
I made several changes to the accordion.

The first one is aimed to fix an issue I just discovered. It appears that you used to always set the `max-height` to a high pixel value so you didn't have to know the height of each item. If it works for when an item expands, it doesn't when an item collapses. Let me explain: when an item is collapsing, the animation progressively reduces the `max-height` towards 0, but as the value is higher than the real height, during a few milliseconds, nothing happens on screen. To remove this kind of delay, we must use the real height of the items.
To do so, I've changed most of the component so it follows these steps:

1. For the first render, the items of the accordion are all shown as expanded.
2. While expanded, we can measure their heights and store them inside `this.state.maxHeights`.
3. We render again but with all the `max-height` to `0`.
4. Each time an item is expanded, its `max-height` is set to the height of its content.

The bad news is that if the window is resized (and thus the content of the item gets a new height), the calculation we made at first isn't accurate anymore. _I did not address this in this PR. I don't know if it's necessary either._

Another change I made prevents the screen to be scrolled at an incoherent position (because an item is collapsing at the same time for example). The change consists in using a double rAF (`requestAnimationFrame`) to ensure the browser updated the layout before we can measure the scroll position of the clicked item. Additionally, the screen is scrolled as soon as possible: sometimes we wait for another item to collapse, sometimes we don't because we don't need to.

Finally, I restored the margin between the content and the title of an item.

**NOTE:** I didn't fix the issue where the clicked item could move up outside of the screen (as explained in the previous comment). I think it will be hard to fix and noticed some other big websites also have it (Barclays for example).

hectoruch:
All is great on FAQ page, but on definitions page where are an accordion does not work well on iPad and Mobile. @clementprdhomme

nerik:
@clementprdhomme Thanks for the fix. I was aware of this issue but thought it was good enough at the time. I guess it's always a tradeoff between time and code simplicity vs product polish :]

nerik:
@clementprdhomme You might want to join the fun here
http://stackoverflow.com/questions/3508605/how-can-i-transition-height-0-to-height-auto-using-css

tiagojsag:
I just tested this. Do the transitions a bit faster. They look nice the first time around, but after that it's just time wasted. Other than that, looks functionally ok

clementprdhomme:
@hectoruch Can you give me more details about the issue? I can't reproduce it on iPhone and iPad.

clementprdhomme:
Thank you @hectoruch for your comment.

I've realised the algorithm used to compute the position of the clicked item was wrong because it would be relative to the first parent with relative position. I've fixed it.

Additionally, I've noticed that, because the entries are loaded asynchronously on the Definitions page, we can't compute the heights of the items at instantiation, I thus made other changes so the animations work again.

Finally, the `autoscroll` prop was broken. It's back.

Can you check everything's great?

hectoruch:
Tested on:
- Firefox
- Chrome
- Edge
- Internet Explorer 11
- IOS
- Android
- Safari

Works great! @clementprdhomme

clementprdhomme:
Ok, so I'll merge it (finally) and we'll see the improvements we can make in the future with the feedback.

PR #267, opened 2016-09-06
Fix/tooltip again

nerik:
https://www.pivotaltracker.com/story/show/129783519

### DIFF
--- a/app/src/components/Shared/ToolTip.jsx
+++ b/app/src/components/Shared/ToolTip.jsx
@@ -18,7 +18,7 @@ class ToolTip extends Component {
 
   onClick() {
     this.setState({
-      visible: true
+      visible: !this.state.visible
     });
   }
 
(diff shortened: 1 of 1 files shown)


clementprdhomme:
* I don't know if it's just on iOS, but when closing once the tooltip, and opening it again, I can't close it anymore.
* Could we just close it by touching somewhere else? It would be like attaching an event to the body or the document.

nerik:
Sorry, I was home and couldn't test this on an actual device, looking at this now.

clementprdhomme:
@hectoruch Could you review that?

On desktop, the tooltip should appear as if we were using the CSS's hover state. On Mobile, the tooltip must appear on click, disappear if clicking again (but not on the tooltip itself) or if clicking elsewhere on the page.

hectoruch:
Tested on:

- Firefox
- Chrome
- Edge
- Internet Explorer 11
- Safari
- IOS
- Android

Works great! Ready to merge.

PR #272, opened 2016-09-07
Removes the horizontal scroll from the homepage

clementprdhomme:
This PR fixes an issue where the homepage could be horizontally scrolled on mobile devices (at least iOS) because the width of a section was fixed to dimension larger than the permitted one.

[Pivotal Task](https://www.pivotaltracker.com/story/show/129919141)


hectoruch:
Test on IOS and Android.
Work great!

PR #274, opened 2016-09-07
update responsive styles for small footer and logic

hectoruch:
Task => https://www.pivotaltracker.com/story/show/129831571

There are elements that not exist on zeplin. 
We are redesign small footer on this task => https://www.pivotaltracker.com/story/show/129919439

image

### DIFF
--- a/app/src/components/Shared/FooterMini.jsx
+++ b/app/src/components/Shared/FooterMini.jsx
@@ -35,8 +35,10 @@ class FooterMini extends Component {
         <div
           style={{
             position: 'absolute',
-            bottom: '38px',
-            width: '100%'
+            bottom: '68px',
+            width: '100%',
+            height: 'calc(100% - 68px - 95px)',
+            overflow: 'auto'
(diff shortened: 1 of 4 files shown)


clementprdhomme:
I would like you to change some things here:
* The light blue "footer" is showing on desktop. I'm not sure that's intended.
* Can you add `-webkit-overflow-scrolling: touch;` inside the scrollable footer?
* Could you extract the CSS from the component into a CSS file?

clementprdhomme:
Thanks @hectoruch, everything's great!

Nevertheless, I want to you to ask Dani to see if we could change the alignment of the two options "Show Footer" and "Support" so they are aligned with the grid. The reason is that, when "Show Footer" turns into "Hide Footer", because their length are different an you use `display: flex;` with `justify-content: space-around;`, the "Support" link is moved by a few pixels. By using `justify-content: space-between;`, we could solve the issue, but the two options needs a padding around.

PR #276, opened 2016-09-07
Update responsive styles

hectoruch:
Update:
- Modal window.
- Articles and Publication titles.
- Video width on tutorial page.

Task => https://www.pivotaltracker.com/story/show/129945125

### DIFF
--- a/app/styles/components/shared/c-modal.scss
+++ b/app/styles/components/shared/c-modal.scss
@@ -13,10 +13,12 @@
   z-index: 1000;
 
   > .content {
+    -webkit-overflow-scrolling: touch;
     background-color: $color-2;
     border: 1px solid rgba($color-3, .3);
     color: $color-4;
-    height: initial;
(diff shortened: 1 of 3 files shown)


clementprdhomme:
Everything's great. Just check with Dani if he doesn't want the modal to go nearly fullscreen on mobile.

app/styles/components/shared/c-modal.scss
```diff
@@ -16,16 +16,14 @@
     background-color: $color-2;
     border: 1px solid rgba($color-3, .3);
     color: $color-4;
-    height: calc(100% - 60px);
-    max-width: 860px;
+    height: initial;
```
clementprdhomme: We don't need this property.

But, we definitely need a `max-height` with an `overflow-y: auto;` set. Also, don't forget the `-webkit-overflow-scrolling: touch;`.

PR #277, opened 2016-09-07
Update suppor form and use modal component for it

hectoruch:
Task => https://www.pivotaltracker.com/story/show/129936447


hectoruch:
@clementprdhomme 
I need to merge that one and https://github.com/Vizzuality/GlobalFishingWatch/pull/276 because they have similars designs and then improve responsive styles.

clementprdhomme:
Not sure that the layout on mobile should be this one. Can you see with Dani if the modal should touch the edges of the screen?
Layout on mobile

Also for windows around 500px wide, the close button is cut:
The close button is cut

hectoruch:
Fix @clementprdhomme

PR #279, opened 2016-09-08
Update images and styles for research and partner pages

hectoruch:
Task => https://www.pivotaltracker.com/story/show/129960153


clementprdhomme:
Some logos are barely readable, we may be increase their size and decrease the one of the bigger. Get in touch with Dani/@tiagojsag to see what we can do.

PR #287, opened 2016-09-12
Update responsive footer with new styles

hectoruch:
Task => https://www.pivotaltracker.com/story/show/130174351


clementprdhomme:
I think the styles of the line should be discussed with Dani. I think it's too bright now. Can you quickly check with him?

hectoruch:
Fixed! @clementprdhomme

PR #288, opened 2016-09-12
Change logo width on sliders

hectoruch:
Change LDF logo width when change the slider.

Task=>https://www.pivotaltracker.com/story/show/130174367

### DIFF
--- a/app/src/components/Home/CoverPage.jsx
+++ b/app/src/components/Home/CoverPage.jsx
@@ -39,7 +39,9 @@ class CoverPage extends Component {
       if (!this.state.autoPlaySlider) {
         return;
       }
-      this.setState({ autoPlaySlider: false });
+      this.setState({
+        autoPlaySlider: false
+      });
     });
     window.addEventListener('resize', this.handleResize);
   }
@@ -156,8 +158,8 @@ class CoverPage extends Component {
     const sliderBackground = sliderBackgrounds.length ? sliderBackgrounds[this.state.currentSlider] : sliderBackground1;
     const sliderAttribution = (sliderBackgrounds.length ?
       sliderAttributions[this.state.currentSlider] : '© OCEANA / Juan Cuetos');
-
     return (
+
       <div
         className={classnames(CoverPageStyle['c-cover-page'], CoverPageStyle['-is-home'])}
         style={{ backgroundImage: `url(${sliderBackground})` }}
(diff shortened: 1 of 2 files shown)


app/src/components/Home/CoverPage.jsx
```diff
@@ -172,8 +189,12 @@ class CoverPage extends Component {
             </div>
             <div className={CoverPageStyle['footer-header']}>
               <div className={CoverPageStyle['contain-ldf']}>
-                <span className={CoverPageStyle['brought-text']}>Brought to you by:</span>
-                <img className={CoverPageStyle['ldf-logo']} src={LogoLDF} alt="logo"></img>
+                <img
+                  className={classnames(CoverPageStyle['ldf-logo'], this.state.windowWidth <= 768 ? 'small-ldf' : null)}
```
clementprdhomme: You could better use this syntax:
```javascript
className({
  [CoverPageStyle['ldf-logo']]: true,
  'small-ldf': this.state.windowWidth <= 768
})
```

app/styles/components/c-cover-page.scss
```diff
@@ -222,19 +231,28 @@
       margin-top: 15px;
       width: 140px;
     }
+
+    :global .small-ldf {
```
clementprdhomme: Do we need this style to be global?

app/src/components/Home/CoverPage.jsx
```diff
@@ -122,6 +131,13 @@ class CoverPage extends Component {
       beforeChange: (currentSlider, nextSlider) => {
         this.onSliderChange(nextSlider);
       },
+      afterChange: this.state.windowWidth <= 768 ? () => {
```
clementprdhomme: Two comments here. First, I would prefer to have the condition inside the callback than having a tertiary operator returning or a function or `null`. Second, I think you're way of determining if "Leo"' slide is active is wrong: the `afterChange` callback receives the current active slide. I would thus save in a variable the index of the slide and just check if the current one matches it.

app/src/components/Home/CoverPage.jsx
```diff
@@ -34,12 +35,20 @@ class CoverPage extends Component {
     this.props.getCoverPageEntries();
   }
 
+  // componentWillReceiveProps() {
```
clementprdhomme: If not used, just remove it. We don't want to introduce extra code.

app/src/components/Home/CoverPage.jsx
```diff
@@ -24,7 +24,8 @@ class CoverPage extends Component {
       currentSlider: 0,
       autoPlaySlider: true,
       speedPlaySlider: 10000,
-      windowWidth: window.innerWidth
+      windowWidth: window.innerWidth,
+      leoTrue: false
```
clementprdhomme: This is never used.

clementprdhomme:
I think your PR is wrong. The task doesn't specify that the foundation's logo has to be bigger on "Leo"'s slide.

hectoruch:
I just fix all @clementprdhomme

clementprdhomme:
@hectoruch I removed the useless refs and squashed the commits into one so we don't have useless commits.

PR #294, opened 2016-09-13
Map perfs

nerik:
This PR focuses only on the vessels layer and does the following:
- makes initial post-tile-loading logic ~10 times faster (5s to 500ms). The bottleneck is clearly now tile data loading times, which is something we don't have control over
- makes frame rendering ~5 times faster, which gets us close to 30 fps displaying 100 days worth of vessel points (on my more than average MB pro; Chrome). Note: to achieve this I simplified greatly the drawing sequence, ie all points have the same fill style (opacity and color). Basically, draw-draw-draw-fill, instead of draw-fill-draw-fill-draw-fill.
  - We'll need to see with the client if this is really needed (the question is: can you really make a difference between a .5 and a .7 opacity vessel point?)
  - if this is a requirement, more logic will be needed to group points in opacity buckets (doing draw-draw-draw-fill for each opacity bucket, ie .5, .75 and 1 for example). This will need more work and will have a cost on performance (but it will still be faster than before)
  - points are now drawn with circles instead of little squares, allowing us to easily encode a value in the circle diameter/area
- removes and cleaned up lot of code, plus some stability fixes (tiles not updating, etc)

Next steps:

- revise track layer:
  - logic has to be revised to work with the new time indexes used in vessel layer (currently track layers are broken)
  - optimize performance
  - work on styling (circles should look the same as in the vessel layer, work on colors, line style, etc). check with Dani
  - show tracks on hover as in the prototype?
- vessel layer render style improvements:
  - experiment with different blend modes and check performance costs
  - check with Dani
- We still have gaps/borders between tiles which is due to 2 things:
  - something is wrong with the data (a tile full of points is a neighbour of a tile almost empty). This is due to inconsistencies in the data. Make sure it's not on our side, if not check with the client
  - tile borders are visible because a point drawn too close to the tile limit it will get cut off. The only fix is to use a full screen canvas instead of tiles, but this is a huge change and I'm not sure how it would perform. Check with David Inga
- Fix timeline glitches and figure out outer bounds thing.

### DIFF
--- a/app/src/components/Layers/CanvasLayer.jsx
+++ b/app/src/components/Layers/CanvasLayer.jsx
@@ -11,7 +11,6 @@ class CanvasLayer {
     this.tileSize = new google.maps.Size(256, 256);
     this.options = _.extend({}, this.defaults, this.options || {});
     this.visible = false;
-    this.filters = filters;
     this.token = token;
     this.setVesselTransparency(vesselTransparency);
     this.setVesselColor(vesselColor);
@@ -19,10 +18,12 @@ class CanvasLayer {
     this.outerStartDate = filters.startDate;
     this.outerStartDateOffset = this._getTimeAtPrecision(this.outerStartDate);
     this.outerEndDate = filters.endDate;
-    this.innerStartDate = filters.timelineInnerExtent[0];
-    this.innerEndDate = filters.timelineInnerExtent[1];
+    this.innerStartDate = filters.timelineInnerExtent[0]; // deprecated
+    this.innerEndDate = filters.timelineInnerExtent[1]; // deprecated
     this.currentInnerStartIndex = this._getOffsetedTimeAtPrecision(this.innerStartDate.getTime());
     this.currentInnerEndIndex = this._getOffsetedTimeAtPrecision(this.innerEndDate.getTime());
+
+    this._setFlag(filters);
 
     if (visible) {
       this.show();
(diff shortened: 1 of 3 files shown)


app/src/components/Layers/CanvasLayer.jsx
```diff
@@ -131,6 +143,39 @@ class CanvasLayer {
...
+    this.playbackData.forEach(canvasPlaybackData => {
+      const canvas = canvasPlaybackData.canvas;
+      const rect = canvas.getBoundingClientRect();
+      if (x > rect.left && y > rect.top && x <= rect.right && y <= rect.bottom) {
+        return this._getCanvasVesselAt(canvas, x - rect.left, y - rect.top);
+      }
+    });
+  }
+
+  _getCanvasVesselAt(canvas, x, y) {
```
nerik: Yeah sorry, I was in the middle of this. I'll remove the last commit from the PR, merge, and open another WIP one.

tiagojsag:
There's a lot of code here, so I'm not going to go into details on all things. But one thing I did notice and that sort of worries me is that we seem to be drawing only canvas pixel per point. This is a quick performance win, but one we probably won't have the luxury of making in the final implementation. We need to visually represent the points in a way that conveys their differences (somehow represent weight and sigma). That is part of both the current beta map and our designs for our map, so simplifying this is something we will not be able to do, and might give us a false sense of performance improvement here.

I'm not saying this is not an improvement, I'm sure it is (although I didn't crunch the numbers myself). What I am saying is that we need to match the implementation with the design, and both meeting half way means we need to save some performance room to accomodate whatever fancy point representation we will need to have.

nerik:
Yes, absolutely aware of this. 
We'll need to play with canvas blend modes and the bucket strategy I mentioned in the PR desc to allow for more visual richness.
Both will come at a perf cost, so more work needs to be done in parallel to keep on improving frame cost - I have a few ideas, and will pick @davidsingal brains.

PR #303, opened 2016-09-13
Update quotes styles on project an research pages

hectoruch:
Task => https://www.pivotaltracker.com/story/show/130284105


clementprdhomme:
Can you remove the `span`? You don't use them. Also, don't forget to remove the empty rule in the CSS.

hectoruch:
Fixed! @clementprdhomme

PR #308, opened 2016-09-13
fix error when user click background

hectoruch:
Task => https://www.pivotaltracker.com/story/show/130301419


clementprdhomme:
First, the code for that is already implemented (method `onClickOverlay`).

Second, the click event bubbles up in the tree, so if you click inside the modal, it will close (unintentionally) the modal. That's why the current method makes a check before calling the prop.

Where have you seen the click on the overlay not working?

PR #309, opened 2016-09-13
update styles for small devices

hectoruch:
Task => https://www.pivotaltracker.com/story/show/130299223


clementprdhomme:
I just noticed the structure of the HTML is wrong. You shouldn't place a `div` within a `ul`.

Also, I don't like the way you positioned the second div, because it totally depends on the length of the first one. If we decide to change its content, we would need to change the CSS too.

Here is a really simple example of how you could handle it: http://jsbin.com/cevuciguwa/edit?html,css,output

In addition, could you change the word "CartoDB" to "CARTO" to follow their new brand name?

PR #310, opened 2016-09-14
Fix close button on mesage after support contact

hectoruch:
Task => https://www.pivotaltracker.com/story/show/130350287

### DIFF
--- a/app/src/actions/contact.js
+++ b/app/src/actions/contact.js
@@ -17,7 +17,9 @@ export function submitForm(data, endpoint) {
       }
       dispatch({
         type: FORM_RESPONSE,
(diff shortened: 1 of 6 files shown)


clementprdhomme:
To prevent this to happen again, can you add `isRequired` to the definition of the prop in `SupportForm.jsx`?

tiagojsag:
- [ ] make the select field required
- [ ] submitting multiple requests after the first is broken (requests are sent, but the interface is not updated)
- [ ] Different browsers have different styles

tiagojsag:
functionally ok, <select > on FF has broken styling

PR #326, opened 2016-09-20
update logic and styles for tooltip and accordion

hectoruch:
Task => https://www.pivotaltracker.com/story/show/130359813

### DIFF
--- a/app/src/components/Shared/ToolTip.jsx
+++ b/app/src/components/Shared/ToolTip.jsx
@@ -56,6 +56,13 @@ class ToolTip extends Component {
   }
 
   showToolTip() {
+    // The height of tooltip is 80px -> 40px
+    // The width of tooltip is 200px
+
+    const tooltipWidth = 200;
+    const tooltipHeight = 80;
+
+
     const bounds = this.refs.info.getBoundingClientRect();
     const left = bounds.left; // Horizontal position relative to the screen
     const height = bounds.height; // Height icon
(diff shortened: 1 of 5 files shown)


tiagojsag:
clean up your code, please. I see a lot of commented CSS

app/src/components/Shared/ToolTip.jsx
```diff
@@ -57,34 +57,42 @@ class ToolTip extends Component {
...
-    // Position relative to the abbr element
-    const offsetTop = this.refs.info.offsetTop;
-
-    // If true, the arrow is on the right of the tooltip
+    const height = bounds.height; // Height icon
+    const width = bounds.width; // Width icon
+    const offset = 10; // Distance between tooltip and icon
+
+    const top = bounds.top + window.scrollY + height + offset; // Calc position top normal tooltip
+    const topRight = bounds.top + window.scrollY - 40 + (height / 2); // Calc position top tooltip with arrow at right
```
clementprdhomme: Can you explain where those 40, 100 and 200 values come from? Is it possible to have them in a variable?

app/src/components/Shared/ToolTip.jsx
```diff
@@ -130,28 +139,30 @@ class ToolTip extends Component {
...
-        className={classnames(ToolTipStyle['c-tooltip-info'], ToolTipStyle[`-${this.props.iconColor || 'gray'}`])}
-        onClick={e => this.onClick(e)}
-        onMouseEnter={() => this.onMouseEnter()}
-        onMouseLeave={() => this.onMouseLeave()}
-        ref="el"
-      >
-        <span
-          className={ToolTipStyle['c-tooltip-info-title']}
-
+      <span>
```
clementprdhomme: Why is there an extra `span` here?

app/styles/components/c-tooltip-info.scss
```diff
@@ -4,62 +4,62 @@
...
-      content: '';
-      height: 0;
-      left: 95px;
-      position: absolute;
-      right: 0;
-      top: -4px;
-      transform: scale(1, -1) rotate(-45deg);
-      width: 0;
-    }
+.c-tooltip-info-content {
```
clementprdhomme: Why is this outside `.c-tooltip-info`? I think it should stay inside.

Also, it shouldn't start with `c-` because it's part of the tooltip component. Maybe we should rename `.c-tooltip-info` to `.c-tooltip` so both the JS and the CSS use the same convention. I know it's not something you introduced here but it would make sense.

clementprdhomme:
Outdated PR.

PR #332, opened 2016-10-04
Vessels layer (canvas only)

nerik:


nerik:
We had a nice time together, canvas. Ciao.

PR #333, opened 2016-10-07
Change content and styles

hectoruch:


tiagojsag:
- [ ] make the social_image smaller. right now it's 2k px wide and 1MB in size

PR #336, opened 2016-10-19
implemented map filters/controls

andresgnlez:
This PR begins the rework of the map controls. It includes:
- Added several SVG icons
- Rework on many SASS files and creation of others accord to the component architecture.
- Updated VesselInfo logic (tooltip). Now its visibility can be switched and position changed.
- Removed old / unnecessary functionalities.
- I've modified the SASS-lint rule about vendor prefixes. Why? I think we all know we shouldn't use them but, unfortunately, sometimes we have to declare them and deal with the warning anyway. That's why I found it pointless.

TO-DO:
- Still need many icons as I indicate in [this task](https://www.pivotaltracker.com/story/show/131872351).
- We are still having troubles adding elements to accordion headers. We need to think about it. The previous solution is limited and only works for a case.
- Not all functionalities/designs are defined. We should talk about this.

Sorry about the PR's size: it's a rebase of the work done during these last 3 days. I would appreciate some feedback (specially the Redux part, I am still a beginner).


app/src/components/Map/ControlPanel.jsx
```diff
@@ -1,21 +1,33 @@
...
+import FiltersPanel from './FiltersPanel';
 import LayerPanel from '../../containers/Map/LayerPanel';
 import SearchPanel from '../../containers/Map/SearchPanel';
 import controlPanelStyle from '../../../styles/components/c-control_panel.scss';
 import { Accordion, AccordionItem } from 'react-sanfona';
 
 class ControlPanel extends Component {
 
+  focusSearch() {
+    document.querySelector('#search-vessels').focus();
```
clementprdhomme: Use a `ref` here.

app/src/components/Map/SearchPanel.jsx
```diff
@@ -1,84 +1,108 @@
...
 
 class SearchPanel extends Component {
 
   constructor(props) {
     super(props);
 
     this.state = {
-      visibleAdvancedSearch: false
+      isEmpty: true,
+      search: this.props.search
```
clementprdhomme: What's the reason why you copy the prop in the state?

app/src/components/Map/SearchPanel.jsx
```diff
@@ -1,84 +1,108 @@
...
     };
+  }
 
-    this.toggleVisibleAdvancedSearch = this.toggleVisibleAdvancedSearch.bind(this);
-    this.onSearchInputChange = this.onSearchInputChange.bind(this);
+  componentWillReceiveProps(nextProps) {
+    if (!!nextProps.search) {
+      Object.assign(this.state.search, nextProps.search);
```
clementprdhomme: Nope, nope, nope. You're already mutating the state here. Because you only want to update the `search` property, just do that instead:
```javascript
this.setState({ search: nextProps.search });
```

app/src/components/Map/SearchPanel.jsx
```diff
@@ -1,84 +1,108 @@
...
 
   onSearchInputChange(event) {
-    this.props.getSearchResults(event.target.value);
-  }
+    this.keyword = event.target.value;
 
-  toggleVisibleAdvancedSearch() {
-    this.setState({
-      visibleAdvancedSearch: !this.state.visibleAdvancedSearch
+    Object.assign(this.state, {
```
clementprdhomme: Same here. You can just read the state, not mutate it.

app/src/components/Map/SearchPanel.jsx
```diff
@@ -1,84 +1,108 @@
...
     return (
-      <div>
+      <div className={searchPanelStyles['c-search-panel']}>
         <input
-          onChange={this.onSearchInputChange}
-          className={controlPanelStyle.input_accordion}
-          placeholder="Type your search query"
+          id="search-vessels"
+          type="text"
+          onChange={(e) => this.onSearchInputChange(e)}
```
clementprdhomme: I think we'll generate way too much queries with that. Can't we have a search button or at least debounce the call?

app/src/components/Map/SearchResult.jsx
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from 'react';
+
+import hightlightWord from '../../util/highlightWord';
+import searchPanelStyles from '../../../styles/components/map/c-search-panel.scss';
+
+class SearchResult extends Component {
+
+  onDrawVessel(event) {
+    const target = event.target.closest('li');
```
clementprdhomme: Better use a `ref` if you can.

app/src/components/Map/VesselInfoPanel.jsx
```diff
@@ -1,86 +1,151 @@
...
+    Object.assign(this.state, nextProps);
+    this.setState(this.state);
+  }
 
-    let flag = '';
-    if (this.props.vesselInfo.flag) {
-      flag = iso3311a2.getCountry(this.props.vesselInfo.flag);
-    }
+  setVars() {
+    this.el = document.querySelector('#vesselBox');
```
clementprdhomme: Use a `ref` please.

app/src/components/Map/VesselInfoPanel.jsx
```diff
@@ -1,86 +1,151 @@
...
     }
 
-    let marineTrafficLink = (
-      <a
-        href={`http://www.marinetraffic.com/en/ais/details/ships/mmsi:${this.props.vesselInfo.mmsi}`}
-        target="_blank"
-      >
-        here
-      </a>);
+    Object.assign(this.state, nextProps);
```
clementprdhomme: You're mutating the state. Really bad 👋🏼.

app/src/components/Map/VesselInfoPanel.jsx
```diff
@@ -1,86 +1,151 @@
...
-    if (this.props.vesselInfo.timestamp) {
-      let humanDate = new Date(this.props.vesselInfo.timestamp).toISOString().slice(0, 10);
-      extraInfo.push(
-        <p key="vesselPanelDate">Date: <span id="vesselPanelDate">{humanDate}</span></p>
-      );
+  setVesselPosition(position) {
+    const top = position.top;
+    const left = position.left;
+
+    this.el.setAttribute('style', `top: ${top + 30}px; left: ${left - 25}px`);
```
clementprdhomme: Should be done in a render otherwise on the next render we could loose the position.

app/src/util/highlightWord.js
```diff
@@ -0,0 +1,6 @@
+export default (strReplace, str, styleClass) => {
```
clementprdhomme: Do we really need this function to be apart in its separate file? If it's used in just one component, let's put it in the component.

app/styles/_settings.scss
```diff
@@ -20,15 +20,18 @@ $duration-picker-height: 24px;
 
 // Font sizes
 $font-size-tiny: 10px;
-$font-size-x-small: 12px;
+$font-size-xxs-small: 12px;
```
clementprdhomme: We're starting to have to much font sizes. Can't you use the closest font-size every time to achieve a similar design?

app/styles/_settings.scss
```diff
@@ -79,3 +85,7 @@ $mq-mobile-small: 'screen and (min-width: #{$breakpoint-mobile-small})';
 $mq-mobile-large: 'screen and (min-width: #{$breakpoint-mobile-large})';
 $mq-tablet: 'screen and (min-width: #{$breakpoint-tablet})';
 $mq-laptop: 'screen and (min-width: #{$breakpoint-laptop})';
+
+// miscellaneous
```
clementprdhomme: This should be with the other colours.

app/styles/components/map/c-layer-panel.scss
```diff
@@ -0,0 +1,68 @@
...
+
+      display: flex;
+      justify-content: space-between;
+
+      &:not(:first-child) {
+        margin: 12px 0 0;
+      }
+
+      .switcher {
+        -webkit-appearance: none;
```
clementprdhomme: Don't we have `autoprefixer` doing it for us?

app/styles/components/map/c-layer-panel.scss
```diff
@@ -0,0 +1,68 @@
...
+        -webkit-transition: .15s;
```
clementprdhomme: You don't need any prefix here.

app/styles/components/map/c-layer-panel.scss
```diff
@@ -0,0 +1,68 @@
...
+        cursor: pointer;
+        height: 10px;
+        margin: 0 8px 0 0;
+        outline: 0;
+        padding: 0 10px 0 0;
+        position: relative;
+        width: 20px;
+
+        &::after {
+          background: $color-26 -webkit-linear-gradient($hsla-end, $hsla-begin);
```
clementprdhomme: We don't need any prefix here. Can you use the same syntax as you did on line 21?

tiagojsag:
@andresgnlez we should probably target "mapster" instead of "develop" on this PR cc @nerik

clementprdhomme:
I did some changes to the PR so we can merge it. I'll add tasks to Pivotal regarding what's missing or to what needs to be improved.

PR #340, opened 2016-11-07
Fix/map viewport

nerik:


tiagojsag:
what is the issue this PR fixes?

nerik:
See commit descriptions.

PR #341, opened 2016-11-07
Workspace

andresgnlez:


clementprdhomme:
Some code already exists to retrieve the workspace ID from the URL, so there's no need to create a new helper. I'll push changes to the workspace management in a new branch.

PR #345, opened 2016-11-07
Flag filter

andresgnlez:
`Filter Panel` is connected with `Map` and sending the flag value once its chosen. But not filtering by flag in the map.


andresgnlez:
Sorry for the commit's name. It took this one from the rebase and I couldn't change it because git hooks doesn't allow it... (cof, cof Erik). Anyway, this commit connects filter panel with map, remove settings panel and add many styles.

The flag filter is being sended to the map, but, for some reason, it's not applied in the current layer, maybe @nerik can help with this.

PR #352, opened 2016-11-15
layer transparency control

andresgnlez:


andresgnlez:
I think tis PR can be merged. There are many improvements to perform (https://www.pivotaltracker.com/story/show/134422291) but we can do them apart.

tiagojsag:
- Spacing between layers is too big
- Would be nice to hace the opacity control auto-hide when you disable a layer

andresgnlez:
Since last commit, the behavior of the opacity control is:

- The opacity control is closed automatically if the user disables the layer.
- If the layer is disabled, the user can open the opacity control and selecting an opacity value the layer will be enabled automatically.

andresgnlez:
@tiagojsag let's focus on behavior by the moment please. These kind of details can be added to a Pivotal task and we can review them at any moment. But this shouldn't stop this PR to be merged or whatever.

PR #353, opened 2016-11-15
Fixed error related with vessel layers

andresgnlez:
Pivotal task: https://www.pivotaltracker.com/story/show/134395665

The error was produced because `ClusterAnimation`-typed layers were trying to access to the method `isVisible` of CARTODB layers.

Also, the property `addedLayers ` of `this.state` was initially an array. Now it's an object. Otherwise, when the app added layers with methods like `addCartoBasemap` or `addCartoLayer`, instead of pushing the layer into the array, the layer was setted as a property of the array, which can be really confusing.

### DIFF
--- a/app/src/components/Map.jsx
+++ b/app/src/components/Map.jsx
@@ -24,7 +24,7 @@ class Map extends Component {
     super(props);
     this.state = {
       overlay: null,
-      addedLayers: [],
+      addedLayers: {},
       lastCenter: null,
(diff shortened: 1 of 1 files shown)


tiagojsag:
Looks ok to me appart from the console.info (and the unrelated lint errors)

andresgnlez:
I would keep `console.info` (they don't break lint)... And they are useful to know what's going on.

PR #356, opened 2016-11-16
zoom to track draft

nerik:


tiagojsag:
@nerik should we keep this implementation, or use the info in the header file, from the API ?

nerik:
We can keep this implementation for now.
My concern now is that I cannot get anything else than 404s from the tracks endpoint (both new and old)
Also, most of the seriesgroup have a negative value (not a problem on our side). The two issue might be related

PR #358, opened 2016-11-16
show/hide vessels

nerik:


tiagojsag:
I see a weird effect when showing the layer: the layer is shown, and then reloaded. So, to the end user, it looks like a bug. Can you improve this, reload before showing?

nerik:
Should be fixed now.

PR #359, opened 2016-11-16
basemaps map control implementation

andresgnlez:
TO-DO:
- [ ]  Implement new basemap layers (`Deep Blue` and `High Contrast` layers)
- [ ]  Thumbnails for the above layers.

This PR contains the next stuff:
- Logic to split basemap layers in another independent section. Only one of them can be selected at the same time.
- Initial styles to unify all list in `mapPanelControl`. If you take a look, all lists are quite similar.. And we have styles for every of them so I'll try to unify them step by step.
- Originally the intention was modify the workspace to manage basemap layers in a better way. But this made the oppossite, so we decide to treat basemap layers as another layer more with the difference that just one of them can be applied at the same time. And no modifying workspace we save work to the API because we keep the original format.
- The`satellite` layer is kind of tricky. Actually we don't manage this "layer". We just take it into account in the `state` to manage the rest of basemaps. This is because it's a GoogleMaps layer and it's different and (hard) so.. In a nuthsell: `satellite` basemap is always in the map, and we play with the visibility of the other basemaps.

To be honest, sometimes, there's a point that map doesn't change basemaps anymore... But I think it's a render stuff and not the logic implemented for the basemaps. Feel free to check it out (and tell me if I'm wrong) and tell your thoughts.


tiagojsag:
- A few seconds after I switch from the initial state to the "Deep blue" basemap, the vessel layer seems to change - probably you are drawing the vessel layer twice or something.
- If I switch from "Satellite" to "High contrast" (without touching the "layers" panel), I see the same basemap but with different layers. Seems like "Satellite" is not showing all layers.

andresgnlez:
image

This is what I get when I switch from `Satellite` to `High contrast`, is this what do you mean, @tiagojsag ?

PR #367, opened 2016-11-23
vesselinfo rework

andresgnlez:
Pivotal task: https://www.pivotaltracker.com/story/show/134290589

- [x] Add Marine Traffic link
- [ ] Add country flag


andresgnlez:
We are not able to implement country flags because search's API is not returning any country code or similar.

andresgnlez:
After speak with Enrique, the country code that allows to display the country flag will be implemented in the new vessel info API by next week.

PR #377, opened 2016-12-01
Extract tileset URL to workspace

tiagojsag:


app/src/components/Layers/VesselsLayerTiled.js
```diff
@@ -1,11 +1,12 @@
 /* eslint no-param-reassign: 0 */
 import VesselsTileData from './VesselsTileData';
 
-class CanvasLayer {
-  constructor(map, token, filters, outerStartDateOffset, debug = false) {
+class VesselsLayerTiled {
```
tiagojsag: Renamed this to make full text searches work, hopefully should have no negative impact on things

PR #383, opened 2016-12-02
Timebar features

andresgnlez:
This PR contains:
 - Some responsive styles
 - Adds Settings duration panel.
 - Allows set predefined time ranges


tiagojsag:
`./app/styles/components/c-map.scss
Module build failed: ModuleBuildError: Module build failed: 
  @media #{$mq-desktop} {
        ^
      Undefined variable: "$mq-desktop".
      in /home/tiagojsag/web/GlobalFishingWatch/app/styles/components/c-map.scss (line 25, column 10)
    at DependenciesBlock.onModuleBuildFailed (/home/tiagojsag/web/GlobalFishingWatch/node_modules/webpack-core/lib/NormalModuleMixin.js:315:19)
    at nextLoader (/home/tiagojsag/web/GlobalFishingWatch/node_modules/webpack-core/lib/NormalModuleMixin.js:270:31)
    at /home/tiagojsag/web/GlobalFishingWatch/node_modules/webpack-core/lib/NormalModuleMixin.js:292:15
    at context.callback (/home/tiagojsag/web/GlobalFishingWatch/node_modules/webpack-core/lib/NormalModuleMixin.js:148:14)
    at Object.onRender (/home/tiagojsag/web/GlobalFishingWatch/node_modules/sass-loader/index.js:272:13)
    at /home/tiagojsag/web/GlobalFishingWatch/node_modules/async/lib/async.js:906:35
    at _arrayEach (/home/tiagojsag/web/GlobalFishingWatch/node_modules/async/lib/async.js:85:13)
    at Object.<anonymous> (/home/tiagojsag/web/GlobalFishingWatch/node_modules/async/lib/async.js:898:17)
    at Object.callback (/home/tiagojsag/web/GlobalFishingWatch/node_modules/async/lib/async.js:44:16)
    at options.error (/home/tiagojsag/web/GlobalFishingWatch/node_modules/node-sass/lib/index.js:286:32)
 @ ./app/src/components/MapIFrame.jsx 31:12-57`

tiagojsag:
- we need to improve the design, right now the outer slider handles are too close to the window edge (desktop)
- The "1 month" tooltip seems to be missing some space on the right
- clicking on the new wheel should trigger some action, right now is triggering a click on the map that is under it
- see also https://www.pivotaltracker.com/story/show/132812309

app/src/components/Map/Timebar.jsx
```diff
@@ -467,6 +471,13 @@ class Timebar extends Component {
   }
 
   render() {
+    const dateFormat = 'DD MMM YYYY';
+    const startDateText = isMobile.phone || isMobile.tablet ? ' start' : 'start date';
```
nerik: Isn´t there a way to do that in CSS? Media queries + ::before + content ?

andresgnlez:
With media queries you have to use widths in px. There's no guarantee for us to know what device is the user using this way. With this library, it reads the User Agent so we are sure about the device the user is using.

app/src/components/Map/Timebar.jsx
andresgnlez: With media queries you have to use widths in px. There's no guarantee for us to know what device is the user using this way. With this library, it reads the User Agent so we are sure about the device the user is using.

nerik: Not sure I understand, this changes the text content to adapt to smaller viewports right? So something like this should work just the same:
```
.whatever {
  &::before {
    content: "start";
  }
  @media #{$mq-mobile-large} {
    &::before {
      content: "start date";
    }
  }
}
```

This is a standard responsive design use case, the lib you are using is intended for redirections/change of feature set depending on device which is different.
Or is there something I'm missing?

tiagojsag: The library relies only on user agents? what about fallback in case of weird user agents? We don't want to have to change the code every time a new startup in china/japan/taiwan/etc releases a new phone. @andresgnlez please spoof user agents and see what happens. While I do get your point that user agents can be more informative, they are also more volatile.

tiagojsag:
- Code is broken, missing imports and such. Please review
- On mobile, the outer range handlers should disappear
- Redefine the granularity of the chart legend.

andresgnlez:
I don't want want this PR to grow more so future features about timebar will go in other different PRs.

nerik:
I made the DurationPicker a bit more DRY by using moment durations as single source of truth in Constants (have a look at the docs here http://momentjs.com/docs/#/durations/).

Furthermore there are a few things off:
- duration label is not placed correctly
screen shot 2016-12-16 at 12 51 39
I don't know what the right close button in the designs is for/if we are going to use it. Please check with Dani, without the close button I suggest to have the settings icon and the label centered together.
- the whole bar overflows off screen. I know you want to address that in a later PR, but I suggest at least you have a look at what introduced the issue now.
- the menu should close when you select an item.

andresgnlez:
`the menu should close when you select an item.` <-- this is on https://github.com/Vizzuality/GlobalFishingWatch/pull/421

tiagojsag:
- [ ] the right handle bar does not always show up

PR #386, opened 2016-12-02
Display message when user clicks on a vessel cluster

nerik:
fish-zoom


nerik:
Merge https://github.com/Vizzuality/GlobalFishingWatch/pull/387 first

PR #396, opened 2016-12-06
optimize webpack image processor

tiagojsag:
This shaved off 5 seconds from the build time, but it's still taking 18 seconds on a cold webpack cache


tiagojsag:
@clementprdhomme can you take a look at the webpack config changes, please? should be harmless, but i just want to make sure

clementprdhomme:
It's all good and goes really fast now!

PR #403, opened 2016-12-07
Save basemap to workspace

tiagojsag:


app/src/reducers/map.js
```diff
@@ -24,7 +23,7 @@ import _ from 'lodash';
 import { DEFAULT_VESSEL_COLOR } from 'constants';
 
 const initialState = {
-  active_basemap: 'satellite',
+  activeBasemap: 'satellite',
```
nerik: For the sake of sanity; let's avoid duplicating source of truth and remove this since this is already in the default workspace.

app/src/components/Map.jsx
```diff
@@ -483,10 +484,10 @@ class Map extends Component {
 
     basemaps.filter((b) => b.type === 'Basemap').forEach((basemap) => {
       const urlTemplate = _.template(basemap.url);
-      this.map.mapTypes.set(basemap.title, new google.maps.ImageMapType({
+      this.map.mapTypes.set(basemap.title.replace(/ /g, '_'), new google.maps.ImageMapType({
```
nerik: Not sure why you need to remove whitespace from this, but better use `/\s/`

PR #406, opened 2016-12-07
Improve duration picker behavior

tiagojsag:
with this, the duration picker updates as you drag the inner extents


nerik:
We did the same thing
https://github.com/Vizzuality/GlobalFishingWatch/pull/408/commits/5cfaaa84fb0d994509929868c3c76efa8ce5634c

:(

PR #410, opened 2016-12-12
better vessel selection and user feedback

nerik:
This PR aims at making vessel selection less opaque to the user:
- when clicked area contains more than one point, which happens often, the series and seriesGroups of all points are checked, and if they all match the proper track is selected, instead of arbitrarily selecting the first point.
- if multiple points in the area give different series[group], or if there's only one but it's a cluster, an message is displayed to the user
- if there are no points at all in the area, a message is displayed to the user
- when info endpoint is loading, a loading message is displayed, giving immediate feedback to user click


tiagojsag:
- when you click on a clustered point, then click on the link to zoom in, please hide the message.
- I like the idea of having a loading message, but we should also have something for the actual tiles. this is perhaps worth a discussion with dani

nerik:
#1 is addressed in last commit
For #2, yes, furthermore Skytruth already requested this. This is also needed for vessels layer loading, so I created a separated issue: https://www.pivotaltracker.com/story/show/135945765

nerik:
Please review and merge https://github.com/Vizzuality/GlobalFishingWatch/pull/411 first.

tiagojsag:
- can you hide the box if you zoom in/out in any other way (for example, double click or mouse scroll)

PR #416, opened 2016-12-13
Fix/timebar overflow

nerik:
screen shot 2016-12-13 at 12 44 55

This PR aims at making more obvious to the user that going beyond outer dates is possible.


tiagojsag:
see if you can take a second look at the weird right handler behavior. Perhaps a good night sleep in between will help. 

also, motivational left shark:

PR #418, opened 2016-12-14
Added attributions on map for mobile

andresgnlez:
Pivotal task: https://www.pivotaltracker.com/story/show/136095157


tiagojsag:
image

we should add something under the attribution text. please talk with dani

PR #420, opened 2016-12-15
reponsive zoom controls

andresgnlez:


app/src/components/Map.jsx
```diff
@@ -538,7 +540,7 @@ class Map extends Component {
    * @param event
    */
   changeZoomLevel(event) {
-    const newZoomLevel = (event.target.id === 'zoom_up')
+    const newZoomLevel = (event.currentTarget.id === 'zoom_up')
```
andresgnlez: I changed this because previously we had literals so `target` worked fine. As we have now SVG (HTML element), we need to update this to `currentTarget`, this way the target of the event is always the element who has assigned the event and not its children. Let me know if you have any doubt about this.

PR #423, opened 2016-12-20
Filtered predefined range time based on timebar current range

andresgnlez:
image

This PR filters predefined range times based on the current timebar. In the above picture the timebar range is lesser than a month, that's why in the predefined range times only the options `15 days` and `7 days` are available.

### DIFF
--- a/app/src/components/Map/DurationPicker.jsx
+++ b/app/src/components/Map/DurationPicker.jsx
@@ -80,15 +80,24 @@ class DurationPicker extends Component {
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


andresgnlez:
Should we limit user range selection to 1 day? If not, if the user chooses a global range lesser than 1 day there will not predefined options available.

tiagojsag:
@andresgnlez yes, lets limit it to a certain value. use a constant so we can change it later on, but nothing under 1 day makes sense

PR #426, opened 2016-12-21
Temporal fix for Safari accordion animations

andresgnlez:
Personally I don't like the idea of pointing to a specific branch but as we don't know when the owner of the project will merge it this is the only solution meanwhile. If the branch is merged we'll have to change this value to the previous one again.

With this change, Safari accordion animations are fixed for Safari desktop, iPad and iPhone.

Don't forget to remove your `react-sanfona` folder in `node_modules` and `npm i` to install the new version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "react-router": "^2.4.1",
     "react-router-redux": "^4.0.4",
     "react-router-scroll": "^0.3.2",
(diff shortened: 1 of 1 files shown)


tiagojsag:
branches are very ephemeral (in our repos, we delete them all the time) but commit hashes are virtually indestructible. this ensures a bit more stability

PR #427, opened 2016-12-22
Report Panel implementation

andresgnlez:
Pivotal task: https://www.pivotaltracker.com/story/show/136573549


andresgnlez:
image

There is a glitch with the timebar SVG that doesn't allow the report panel to be clicked on its bottom. Normally, this is a `z-index` problem but in this case it isn't that easy... So I'll have to take a deeper to this.

tiagojsag:
@andresgnlez leave info for @nerik to take a look at, maybe he knows more about this.

andresgnlez:
@nerik not too much to say: the timebar overlaps the panel because of its top margin. Despite of playing with z-index, we need timbar's z-index to be higher than report panel one, to hide it when it translates (open/close). Take a look, if you don't find it out, don't worry, we'll take a look after holidays!

PR #429, opened 2016-12-30
Feature/layers opacity and color

nerik:


nerik:
hue

tiagojsag:
- [ ] disabling a layer toggles up the transparency controls modal.
- [ ] can we add white as a possible color to the vessel layer?

nerik:
Thanks for the fix. I added 2nd item to PT https://www.pivotaltracker.com/story/show/137080295

PR #430, opened 2017-01-04
Feature/report

nerik:


tiagojsag:
Like it, nice job ;)
a few things:

- [ ] once you disable reporting mode, you can still click the polygons and see the popups for that layer
- [ ] I currently see numbers instead of user friendly names on the polygons after clicking
- [ ] Can we replace "polygon(s)" with "region(s)" or something more relatable on the interface?
- [ ] Can we make the polygon itself clickable when reporting, and not just the borders? If not, we should thing of an alternative: make borders thicker, change mouse cursor on hover (bad on mobile) or some other way to improve the experience

nerik:
> once you disable reporting mode, you can still click the polygons and see the popups for that layer
Can't reproduce, could you give me a scenario plz?

tiagojsag:
ok, change the labels and fix the bug, and let's merge this and add the rest to pivotal

PR #431, opened 2017-01-05
Fix filter for Andorra case

tiagojsag:


nerik:
It's `isNaN` that you were looking for.
But having NaN in the first place was weird, so I fixed that and used an explicit `===undefined` (because 0, Andorra's code, is falsy, yippeee JS)

tiagojsag:
I tried the NaN comparison, didn't work. anyway, I trust your js skills more than mine, so as long as the bug is fixed, I'm happy

nerik:
It should, and also, `isInteger` is not supported by IE.

nerik:
Oh and FYI, in JS:
```
NaN === NaN
false
```
:p

PR #432, opened 2017-01-09
Remove map events properly

andresgnlez:


andresgnlez:
When `componentWillUnmount` triggers are removes all events attached to map, once you try to go back to the map nothing happens on it. This "feature" seems to be there always but until now, due to the current error (function not found) this didn't happen.

nerik:
I wasn't aware you couldn't use removeListener directly, thanks.
I'm looking at this now.

nerik:
So it turns out you can remove all listeners at once with `clearInstanceListeners`.
The actual problem, the vessels layer not loading the second time you visit the map, is not related to this. I will fix it in https://github.com/Vizzuality/GlobalFishingWatch/pull/434

andresgnlez:
@nerik That's it. I saw that method but I wasn't very sure to use it because with that your are removing ALL events of the instanced map and IMHO maybe this is not the best place to do that (I would only remove the polygon-related events) but if you find there's no problem with this. Go ahead 😁

PR #436, opened 2017-01-09
Layer Library Modal

andresgnlez:


andresgnlez:
Probably this needs some changes after talking with Tiago, but we'll need this codebase.

app/src/containers/Map.js
```diff
@@ -24,7 +25,8 @@ const mapStateToProps = (state) => ({
   basemaps: state.map.basemaps,
   activeBasemap: state.map.activeBasemap,
   layerModal: state.map.layerModal,
-  supportModal: state.map.supportModal
+  supportModal: state.map.supportModal,
+  layerLibraryModal: state.map.layerLibraryModal
```
nerik: Avoid using "black box" objects like this one. This makes the container harder to read.
Here you just need the bool `state.map.layerLibraryModal.open`

app/src/components/Map/LayerLibrary.jsx
```diff
@@ -0,0 +1,52 @@
...
+class LayerLibrary extends Component {
+
+  render() {
+    if (!this.props.layers) return null;
+
+    const library = [];
+
+    if (this.props.layers.length > 0) {
+      this.props.layers.forEach((layer, i) => {
+        library.push(<LayerItem
```
nerik: I'm not sure it's a good idea to reuse LayerItem. In the end we will stuff this component with `if (!this.props.layerLibraryDisplay)`, callbacks will not be the same, it will become an unreadable mess: because noth "Layers" fulfill a completely different functionality.
In fact I don't quite understand the whole AddLayer flow, the layers here should just be layer titles added to the main layers thing, we need to discuss with Dani.

nerik: But I guess it's ok for now

PR #437, opened 2017-01-10
Feature/per sublayer filtering

nerik:
Flag setting is now per layer, including in the workspace.
Check with Skytruth for the implications of this.
Note: there are no UI elements to filter per layer yet, and the former flag selector is not working anymore for the moment.

### DIFF
--- a/app/src/actions/layers.js
+++ b/app/src/actions/layers.js
@@ -20,6 +20,9 @@ export function initLayers(layers_) {
       } else {
         l.opacity = 1;
       }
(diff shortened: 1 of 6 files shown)


nerik:
gfw-double

nerik:
Task added to PT : https://www.pivotaltracker.com/story/show/137370875
"Heatmap sublayers should be able to share the same tile data"

andresgnlez:
Looks good!

nerik:
🎄

PR #438, opened 2017-01-10
Added z-index system to modal component

andresgnlez:
This system avoid modals overlap each other giving a z-index value to everyone need it and solves problem with the two modals when layer library is being used.


tiagojsag:
seems ok to me. @nerik ?

PR #439, opened 2017-01-11
ACL Implementation 

andresgnlez:


tiagojsag:
@andresgnlez need your help to style something, let's talk tomorrow

app/src/actions/user.js
```diff
@@ -10,6 +12,14 @@ export function setToken(token) {
   };
 }
 
+function getUserData(data) {
+  return data;
```
nerik: It's a no-op, what is it for?

app/src/containers/Map.js
```diff
@@ -28,7 +28,8 @@ const mapStateToProps = (state) => ({
   activeBasemap: state.map.activeBasemap,
   layerModal: state.map.layerModal,
   supportModal: state.map.supportModal,
-  layerLibraryModal: state.map.layerLibraryModal.open
+  layerLibraryModal: state.map.layerLibraryModal.open,
+  userPermissions: state.user.acl
```
nerik: Why calling this `userPermissions` in containers/components, and `acl` in the app state?

nerik:
@tiagojsag just two comments, otherwise lgtm 👍

PR #442, opened 2017-01-11
Workspace compatibility

tiagojsag:


app/src/actions/workspace.js
```diff
@@ -0,0 +1,135 @@
...
+  // TODO this is only used by vesselInfo, but the data is inside a layer
+  // review wit SkyTruth
+  dispatch({
+    type: SET_TILESET_URL, payload: tilesetUrl
+  });
+
+  dispatch(initLayers(workspace.map.layers));
+}
+
+export function processProductionWorkspace(data, dispatch) {
```
nerik: You don't need to export this function

app/src/actions/workspace.js
```diff
@@ -0,0 +1,135 @@
+import { LAYER_TYPES } from 'constants';
+import {
+  SET_LAYERS, SET_ZOOM, SET_CENTER, SET_INNER_TIMELINE_DATES, SET_OUTER_TIMELINE_DATES, SET_BASEMAP, SET_TILESET_URL
+} from 'actions';
+import { initLayers } from 'actions/layers';
+
+export function processNewWorkspace(data, dispatch) {
```
nerik: You don't need to export that function

app/src/actions/workspace.js
```diff
@@ -0,0 +1,135 @@
...
+    fetch(url, {
+      headers: {
+        Authorization: `Bearer ${state.user.token}`
+      }
+    }).then(res => res.json())
+      .then(data => {
+        if ('workspace' in data) {
+          return processNewWorkspace(data, dispatch);
+        }
+        return processProductionWorkspace(data, dispatch);
```
nerik: more like `processLegacyWorkspace` ? The new format is gonna be in production too at some point.

app/src/actions/workspace.js
```diff
@@ -0,0 +1,135 @@
...
+      url = `${MAP_API_ENDPOINT}/v1/workspaces/${ID}`;
+    }
+
+    fetch(url, {
+      headers: {
+        Authorization: `Bearer ${state.user.token}`
+      }
+    }).then(res => res.json())
+      .then(data => {
+        if ('workspace' in data) {
```
nerik: data.workspace !== undefined

nerik: (the in operator can expect unexpectedly with the object's prototype chain, you should use hasOwnProperty, === undefined is a more common convention)

nerik:
IMHO we should not duplicate action dispatches, but rather build a workspace object with the same structure, then dispatch actions at a single place. We will easily forget to add those in both places in the future.

tiagojsag:
that workspace object with the same structure for both is the redux state container.

nerik:
> we should not duplicate action dispatches
was my main point

nerik:
-_-

PR #444, opened 2017-01-12
Moved tileset dispatch to keep consistent layer format

andresgnlez:


nerik:
Could you please add a technical issue in PT to fix that vesselInfo endpoint thing?
Otherwise LGTM 👍

andresgnlez:
Done: https://www.pivotaltracker.com/story/show/137514619

PR #445, opened 2017-01-12
Adding/Removing layers

andresgnlez:


andresgnlez:
https://github.com/Vizzuality/GlobalFishingWatch/blob/mapster/app/src/components/Layers/MapLayers.jsx#L173

The way we update layers trusts the number of layers doesn't change (we ask for the same index in the current and incoming layers)  but this is not true now we can add new layers so new layers can be larger or smaller than current one. I would like to know your thoughts about this, @nerik 

Thanks!

PR #450, opened 2017-01-13
Flag filters UI

andresgnlez:
TODO:

- [x] Change selector literal (ALL => ALL COUNTRIES)
- [x] Change size of selector (less height)
- [x] Make font size smaller
- [x] Make selector arrow clickable 
- [x] Changed map controls  background opacity
- [x] Separate selectors.


tiagojsag:
removing a filter is not reflected on the map. other than that, really good ;)

andresgnlez:
Ops, you are right. Let me add it!

andresgnlez:
`removing a filter is not reflected on the map. other than that, really good ;)`

In the end, @nerik will implement this.

PR #452, opened 2017-01-16
Feature/load tileset metadata

tiagojsag:


tiagojsag:
Loads info from layers:
- max zoom
- tileset max time range

PR #455, opened 2017-01-17
Multiple/pinned tracks

nerik:


nerik:
Tracks pinning + multiple rendering + coloring

This is missing (to be done on a separated PR): 
- the actual hue selector (now it's a button sending a random hue)
- UI styling

andresgnlez:
UI styling and hue selector trigger are in https://github.com/Vizzuality/GlobalFishingWatch/pull/459

app/src/actions/vesselInfo.js
```diff
@@ -122,9 +121,59 @@ sub/seriesgroup=${seriesGroup}/${i}-01-01T00:00:00.000Z,${i + 1}-01-01T00:00:00.
...
+  };
+}
+
+export function toggleActiveVesselPin() {
+  return {
+    type: TOGGLE_ACTIVE_VESSEL_PIN
+  };
+}
+
+export function toggleVesselPin(seriesgroup) {
```
tiagojsag: I don't like toggle functions, as we are always dependent on hidden previous state, which is not always visible. Can you at least add an optional argument to force a certain state in case we want to?

app/src/components/Layers/MapLayers.jsx
```diff
@@ -95,22 +96,30 @@ class MapLayers extends Component {
...
+    if (this.props.vesselTracks.length !== nextProps.vesselTracks.length) {
       return true;
     }
-    if (this.props.vesselTrack.selectedSeries !== nextProps.vesselTrack.selectedSeries) {
+    if (nextProps.vesselTracks.some((vesselTrack, index) =>
+      vesselTrack.hue !== this.props.vesselTracks[index].hue
+    ) === true) {
       return true;
     }
+    // if (this.props.vesselTracks.seriesgroup !== nextProps.vesselTracks.seriesgroup) {
```
tiagojsag: remove?

app/src/components/Layers/MapLayers.jsx
```diff
@@ -95,22 +96,30 @@ class MapLayers extends Component {
...
+    //   return true;
+    // }
     if (this.props.zoom !== nextProps.zoom) {
       return true;
     }
     if (this.props.timelinePaused !== nextProps.timelinePaused) {
       return true;
     }
-    if (nextProps.vesselTrack.selectedSeries && extentChanged(this.props.timelineOverExtent, nextProps.timelineOverExtent)) {
+    if (/* nextProps.vesselTrack.selectedSeries && */extentChanged(this.props.timelineOverExtent, nextProps.timelineOverExtent)) {
```
tiagojsag: remove that commented part?

app/src/actions/vesselInfo.js
nerik: I kind of see what you mean but here SRP + DRY: only the reducer knows about a given pinned track status. If we do a `pinVesselTrack/unpinVesselTrack` action, it means the view has knowledge of this status which is something I'd rather avoid. 

I'm open to suggestions.

TOGGLE_ACTIVE_VESSEL_PIN is a misleading action type though, will fix.

PR #458, opened 2017-01-18
fixed offset issue at high z levels

nerik:
Replaced our hacky coordinates calculation with standard GMaps APIs, which will be slower, but it's done 'offline' (only once after tile loading) so it's barely visible (especially since we are still loading all years at once), but we might want to try to revise that at a later stage

### DIFF
--- a/public/workspace.json
+++ b/public/workspace.json
@@ -38,7 +38,7 @@
         {
           "title": "VESSEL",
           "description": "Fishing effort layer (fr)",
-          "url": "https://api-dot-world-fishing-827.appspot.com/v1/tilesets/801-tileset-nz2-tms",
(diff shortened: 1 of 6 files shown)


public/workspace.json
```diff
@@ -38,7 +38,7 @@
         {
           "title": "VESSEL",
           "description": "Fishing effort layer (fr)",
-          "url": "https://api-dot-world-fishing-827.appspot.com/v1/tilesets/801-tileset-nz2-tms",
+          "url": "https://api-dot-world-fishing-827.appspot.com/v1/tilesets/849-tileset-tms",
```
tiagojsag: is this intentional?

PR #462, opened 2017-01-18
Set default hue value in filters

andresgnlez:


andresgnlez:
I'm not sure if this what you told me about setting default values on reducers. Actually it works, but I want to know if this is what you meant. 

Thanks.

nerik:
Setting a default filter hue (using the layer hue) is already done in the filters action: https://github.com/Vizzuality/GlobalFishingWatch/blob/mapster/app/src/actions/filters.js#L35

So this shouldn't be needed. Sorry if I was unclear.

PR #466, opened 2017-01-19
Rebase/pinned vessels ui

andresgnlez:


nerik:
Nice work. A few issues:
- delete button appears in black
- hue selector is not initialized with the correct hue

nerik:
- edit mode should be toggled off when panel collapses

nerik:
- hue icon should be white when hue selector is opened (as in layers panel)

PR #467, opened 2017-01-19
Search panel redesign

nerik:


app/src/components/Map/SearchPanel.jsx
```diff
@@ -14,7 +15,8 @@ class SearchPanel extends Component {
     super(props);
 
     this.defaults = {
-      characterLimit: 3
+      characterLimit: 3,
+      resultsLimit: 4
```
nerik: Move both variables to constants please (ie SEARC_RESULTS_CHARACTER_LIMIT)

app/src/components/Map/VesselInfoPanel.jsx
```diff
@@ -15,7 +15,7 @@ class VesselInfoPanel extends Component {
     const vesselInfo = this.props.details.find(vessel => vessel.visible === true);
     const status = this.props.detailsStatus;
 
-    if (status === null && vesselInfo === undefined) {
+    if (status === null || vesselInfo === undefined) {
```
andresgnlez: @nerik I had to change this because when you click on a vessel, change the search (keeping the vesselInfo open) and click on a new result the `status` variable was null but not the `vesselInfo` one so it tried to render triggering an error. 

Anyway, take a look, please, and tell me if you find something strange, thanks!

PR #469, opened 2017-01-19
Implemented edition mode for pinned vessels

andresgnlez:
TODO:

- [x] Set action from reducer that updates pinned vessels data.


app/src/components/Map/PinnedTracks.jsx
```diff
@@ -35,6 +32,17 @@ class PinnedTracks extends Component {
...
 
+  onUpdatedItem(updatedVessel, index) {
+    const pinnedVessels = this.props.vessels.filter(vessel => vessel.pinned === true);
+
+    if (pinnedVessels[index] === undefined) return;
+
+    Object.assign(pinnedVessels[index], updatedVessel);
+
+    // WIP
+    // this.props.savePinnedVessels(pinnedVessels);
```
andresgnlez: Here should go the reducer action that updates pinned vessels.

PR #470, opened 2017-01-19
Search Modal

andresgnlez:


tiagojsag:
@andresgnlez I used a react component for pagination, so I needed to change your CSS. please take a look at it, I am not responsible for my own CSS :P

@nerik please do the review

app/src/components/Map/SearchModal.jsx
```diff
@@ -0,0 +1,124 @@
...
+          </div>
+          {searchResults && <ul className={classnames(ResultListStyles['c-result-list'], SearchModalStyles['search-result-list'])}>
+            {searchResults}
+          </ul>
+          }
+        </div>
+        <div className={SearchModalStyles['paginator-container']}>
+          <div
+            className={PaginatorStyles['c-paginator']}
+            style={(!this.props.searching && this.props.entries.length > 0) ? {} : { display: 'none' }}
```
tiagojsag: https://github.com/AdeleD/react-paginate/issues/91

If we do not render the paginator when in "searching" state, then we need to pass it an "initialPage" parameter to set the page (otherwise always shows the first).
If we do this, it triggers the page changed callback, which basically creates an endless loop. So, we can't remove the Paginator, just hide it. A bit hacky, I know.

andresgnlez:
@tiagojsag change page trigger is not working.

app/src/components/Map/SearchModal.jsx
```diff
@@ -1,53 +1,59 @@
...
-
 import SearchIcon from 'babel!svg-react!assets/icons/search-icon.svg?name=SearchIcon';
 import CloseIcon from 'babel!svg-react!assets/icons/close.svg?name=CloseIcon';
 
 class SearchModal extends Component {
 
+  constructor(props) {
+    super(props);
+
+    this.onPageChange = this.onPageChange.bind(this);
```
andresgnlez: No need to bind functions here. You can get rid of the constructor.

app/src/components/Map/SearchModal.jsx
```diff
@@ -1,53 +1,59 @@
...
       }
-    } else if (this.props.keyword.length < SEARCH_QUERY_MINIMUM_LIMIT && this.props.keyword.length > 0) {
+    } else if (this.props.searchTerm.length < SEARCH_QUERY_MINIMUM_LIMIT && this.props.searchTerm.length > 0) {
       searchResults = (
         <li className={SearchModalStyles['status-message']}>
           Type at least {SEARCH_QUERY_MINIMUM_LIMIT} characters
         </li>);
     } else {
-      searchResults = <li className={ResultListStyles['status-message']}>No result</li>;
+      searchResults = <li className={SearchModalStyles['status-message']}>No result</li>;
```
andresgnlez: Why do you change the CSS class?

app/src/components/Map/SearchModal.jsx
```diff
@@ -59,48 +65,60 @@ class SearchModal extends Component {
...
+          <div
+            className={PaginatorStyles['c-paginator']}
+            style={(!this.props.searching && this.props.entries.length > 0) ? {} : { display: 'none' }}
+          >
+            <ReactPaginate
+              previousLabel={<Rhombus />}
+              nextLabel={<Rhombus />}
+              nextClassName={PaginatorStyles.next}
+              previousClassName={PaginatorStyles.previous}
+              breakLabel={<a href="">...</a>}
```
andresgnlez: I think this shouldn't be a link.

tiagojsag: fixed

tiagojsag:
@andresgnlez ready!

PR #476, opened 2017-01-23
Reporting

nerik:


nerik:
report

nerik:
For now only works with the MPA- no take layer, as I prepared it on my Carto account in advance

tiagojsag:
I'm wondering if it makes sense to auto-close the tooltip after you press "add to report". Your opinion @nerik ?

nerik:
@tiagojsag I don't think so, keeping the tooltip opened helps the user understand the "added to report/removed from report" mechanism.
@danielcasius ?

tiagojsag:
yes, I see the value in that. However, as I tested, I ended up closing that modal right after, every time. Since the remove interaction is also duplicated between the tooltip and the layer list on the bottom right, I am not sure if it's worth leaving it open. But yeah @danielcasius can decide for us

danielcasius:
Let's try not closing the tooltip, letting the user know the "Added to report" status. This tooltip should be closed by clicking anywhere on the map, to be a non-intrusive element.

nerik:
@tiagojsag Dani's feedback included, it's ready to be merged

PR #491, opened 2017-01-27
Feature/tilesets headers

nerik:


nerik:
- [ ] debug with real workspace
- [ ] load header when adding layer from directory (https://github.com/Vizzuality/GlobalFishingWatch/pull/491/files#diff-a51b112676f1a79b806e832fe3a97fb5R132)

nerik:
- properly loads all heatmap layers headers
- uses date boundaries from heatmap headers
- fixes all issues related to showing multiple heatmap layers + filters
- modifies the heatmap reduxer so that it has its own representation of active tiles regardless of what heatmap layers are shown
- fixes a critical bug where the heatmap shown was actually not matching the dates in the timebar 😨
- fixes a critical bug where some points would be completely omitted because opacity was equal to NaN 😨
- actually deletes a heatmap layer stage when removing from workspace (still needs to be done for carto and kml layers)

nerik:
btw: this PR changes the default workspace so that it only shows indonesia, then you can add the normal vessel layer
that’s the only way to test it as the indonesia layer is not in the library yet

PR #494, opened 2017-01-31
Pinned vessel changes

tiagojsag:
- save and load hue
- visibility toggle


nerik:
pin

nerik:
I finished pinning and toggling on/off, I'll do hue on another branch.

PR #495, opened 2017-01-31
test iframe page

nerik:
Unless I'm mistaken there's actually nothing to do on our side.


nerik:
Unless there's some kind of interaction between the host and the map within the iframe?

PR #501, opened 2017-02-02
Feature/gl tracks

nerik:


nerik:
Tracks are way faster 🤘

tracks

nerik:
I also tried a new style to highlight the track points while not conflicting with the heatmap too much, but nothing definitive, let's discuss about this.

PR #508, opened 2017-02-06
change time range to fit searched vessel track

nerik:
When searching for a vessel, this will resize the timebar to only show the range covering the vessel tracks.

https://www.pivotaltracker.com/story/show/135521081


app/src/actions/vesselInfo.js
```diff
@@ -13,10 +14,14 @@ import {
   SET_PINNED_VESSEL_TITLE,
   TOGGLE_PINNED_VESSEL_EDIT_MODE,
   SET_RECENT_VESSEL_HISTORY,
-  LOAD_RECENT_VESSEL_HISTORY
+  LOAD_RECENT_VESSEL_HISTORY,
+  TIMELINE_MIN_INNER_EXTENT
```
tiagojsag: did you test this properly? you are importing TIMELINE_MIN_INNER_EXTENT from actions.js, but it's defined in constants.js

PR #509, opened 2017-02-06
lat lon on hover

nerik:
https://www.pivotaltracker.com/story/show/136899151


sorodrigo:
On medium sized screens the latlong tag clashes with the zoom controls. Apart from that perfect 👍 .

PR #513, opened 2017-02-08
Feature: responsive style fixes

sorodrigo:
This PR includes the following:

- [x] Fix app height on mobile devices (it shouldn't be scroll).
- [ ] Review layer lists (in Layers Accordion tab) styles in Safari/iOS. (react-sanfona issue)
- [x] Add overflow hidden to body when modals show up. This way the user can't scroll the app, just the modal.
- [x] Switcher component & making sure its crossbrowser.
- [ ]  Timebar

### DIFF
--- a/app/src/components/Shared/Modal.jsx
+++ b/app/src/components/Shared/Modal.jsx
@@ -43,8 +43,11 @@ class Modal extends React.Component {
 
   render() {
     if (!this.props.opened) {
+      document.body.style.overflow = 'overlay';
       return null;
     }
+
+    document.body.style.overflow = 'hidden';
 
     let closeButton;
     const customStyles = {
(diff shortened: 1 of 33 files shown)


app/src/components/Shared/Modal.jsx
```diff
@@ -43,9 +43,12 @@ class Modal extends React.Component {
 
   render() {
     if (!this.props.opened) {
+      document.body.style.overflow = 'overlay';
```
nerik: State styles should be applied with a class and defined in CSS. The convention is `.-hidden` or `.-expanded` for example.
(not saying it's necessarily better, but it's the convention in this project, let's try to be consistent)

nerik:
Thanks !

PR #514, opened 2017-02-08
Feature/sharks layer

nerik:


nerik:
sharks

nerik:
Zoom close to Quebec to see.
Only rendered as heatmap points for now. Also, points are quite small. This should be ok for a PoC

PR #519, opened 2017-02-10
Fix/hue pickers

nerik:
Merge https://github.com/Vizzuality/GlobalFishingWatch/pull/494 first.

Fixes various issues on the blending tooltips. 
- they were appearing empty on pinned tracks and filters
- they were closed automatically when changing the values on layers
- fill style was not closed on tracks, leading to rendering bugs
- pinned track is set to visible automatically when setting hue (same as layers)
- cleanup and small refacto


nerik:
@sorodrigo plz wait for #494 to be merged before reviewing

PR #520, opened 2017-02-10
country list fixes

nerik:
- move emoji at the end of the country name (keyboard accessibility)
- added a css rule that might fix the white on white country list on linux
@tiagojsag could you test on your machine?


nerik:
screen shot 2017-02-13 at 10 42 59

Visible fallback.

PR #522, opened 2017-02-10
Feature: custom layer url

sorodrigo:
This PR includes the following changes:
- Replaces file field from the custom layer upload form with a url text field.
- Updates the redux flow to use url instead of file.
- Adds request to the api to create the custom layer.
- Adds the id attribute to the created layer object.


app/src/actions/customLayer.js
```diff
@@ -5,24 +5,37 @@ import {
 import { setLayerManagementModalVisibility } from 'actions/map';
 import { addCustomLayer } from 'actions/layers';
 
-export default function uploadLayer(file, name, description) {
-  return (dispatch) => {
+export default function uploadLayer(url, name, description) {
```
nerik: Misleading name, `addCustomLayer` for example.

nerik:
A few things:
- [ ] rename UploadLayer.jsx to a more appropriate name, ie CustomLayer.jsx for example
- [ ] change the modal tab label to 'Custom Layer'
- [ ] apparently the API seems to be happy with any URL. Can you ask Skytruth if they plan to implement some URL validity checking? And if so agree on an error format, and implement error handling on the client side (on another ofc)

Thanks 👍

sorodrigo:
@nerik done!

sorodrigo:
custom-layer-example

nerik:
Yeah!
Have you discussed with Dani about that gray toggle background by default?
Otherwise, LGTM \o/

PR #523, opened 2017-02-13
Fix/blending tooltips

nerik:
Fixes various issues on the blending tooltips. 
- they were appearing empty on pinned tracks and filters
- they were closed automatically when changing the values on layers
- fill style was not closed on tracks, leading to rendering bugs
- pinned track is set to visible automatically when setting hue (same as layers)
- cleanup and small refacto


app/src/components/Layers/TracksLayer.js
```diff
@@ -103,6 +103,7 @@ export default class TracksLayerGL {
       // this.stage.drawCircle(circlePoints.inner.x[i], circlePoints.inner.y[i], circlePoints.inner.radius[i]);
       this.stage.drawCircle(circlePoints.inner.x[i], circlePoints.inner.y[i], 2);
     }
+    this.stage.endFill();
     // // inner range glow
     // this.stage.beginFill(color, 0.1);
```
sorodrigo: Is this code going to be used in the near future? If its old stuff, we should better remove it.

PR #524, opened 2017-02-13
Progressively load tilesets w/ temporalExtents

nerik:
This aims at loading only parts of the tilesets, using only the temporalExtents from the header to match with the currently selected outer range. Currently WIP


nerik:
Making fishing great again by lazy loading time ranges for the heatmap layer. 
With this, data loaded is initially 20% of the whole data instead of 100%

PR #530, opened 2017-02-13
feature: pinned vessels text

sorodrigo:
This PR adds a heading to the pinned vessels accordion.


nerik:
Noice :)

I fixed it, but please keep in mind for later: the scss linter should preferably pass before submitting a PR for review. It's not in the git hook because we thought it would be a bit annoying, but you should be able to set your IDE to automatically highlight SCSS errors/style issues.

PR #531, opened 2017-02-13
feature: resize after load crash

sorodrigo:
This PR address an issue where the map crashed if resized immediately after loading.
- [x] fixes crash when resizing immediately after load.


nerik:
Thanks ! 👍

PR #536, opened 2017-02-14
feature: track hover zoom styles

sorodrigo:
This PR includes the following changes:
- Changes hover thickness and style on depending on zoom level.

### DIFF
--- a/app/src/components/Layers/TracksLayer.js
+++ b/app/src/components/Layers/TracksLayer.js
@@ -33,6 +33,7 @@ export default class TracksLayerGL {
    */
   _drawTrack(data, queriedSeries, hue, drawParams, offsets) {
     const color = hueToRgbHexString(hue);
+    const MIN_ZOOM_LEVEL = 5;
 
     let prevDrawStyle;
     let prevSeries;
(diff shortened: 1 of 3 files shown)


app/src/components/Layers/TracksLayer.js
```diff
@@ -33,6 +33,7 @@ export default class TracksLayerGL {
    */
   _drawTrack(data, queriedSeries, hue, drawParams, offsets) {
     const color = hueToRgbHexString(hue);
+    const MIN_ZOOM_LEVEL = 5;
```
nerik: Could you change the name to give it more context, like `TRACKS_DOTS_STYLE_ZOOM_THRESHOLD` for example, and move it to constants (see https://github.com/Vizzuality/GlobalFishingWatch/blob/develop/app/src/constants.js#L53)

app/src/components/Layers/TracksLayer.js
```diff
@@ -99,18 +104,20 @@ export default class TracksLayerGL {
     this.stage.lineStyle(0);
 
     // inner range center circle
-    this.stage.beginFill('0xFFFFFF', 1);
-    for (let i = 0, circlesLength = circlePoints.inner.x.length; i < circlesLength; i++) {
-      this.stage.drawCircle(circlePoints.inner.x[i], circlePoints.inner.y[i], 2);
-    }
-    this.stage.endFill();
+    if (drawParams.zoom > MIN_ZOOM_LEVEL) {
```
nerik: If drawParams.zoom < MIN_ZOOM_LEVEL, you don't need to collect `circlePoints` in the loop above. Maybe that if should be a bit above (then just check for `circlePoints`'s existence.

sorodrigo:
Ready @nerik!

nerik:
Thanks!

PR #537, opened 2017-02-15
close pinned vessels tooltip

sorodrigo:
This PR fixes the issue where tooltip weren't closing on click outside of them.

### DIFF
--- a/app/src/components/Map/LayerBlendingOptionsTooltip.jsx
+++ b/app/src/components/Map/LayerBlendingOptionsTooltip.jsx
@@ -91,7 +91,7 @@ class LayerOptionsTooltip extends Component {
         <BlendingIcon
           onClick={() => this.props.toggleVisibility()}
           className={classnames(icons['blending-icon'],
-            { [`${icons['-white']}`]: this.props.showBlending })}
+            { [`${icons['-white']}`]: this.props.visible })}
         />
         {this.props.visible &&
         <div
@@ -137,8 +137,7 @@ LayerOptionsTooltip.propTypes = {
   onChangeOpacity: React.PropTypes.func,
   isReverse: React.PropTypes.bool,
   toggleVisibility: React.PropTypes.func,
-  visible: React.PropTypes.bool,
-  showBlending: React.PropTypes.func
+  visible: React.PropTypes.bool
 };
 
 export default LayerOptionsTooltip;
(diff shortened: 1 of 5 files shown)


app/src/components/Map/LayerBlendingOptionsTooltip.jsx
```diff
@@ -46,6 +47,23 @@ class LayerOptionsTooltip extends Component {
       opacityRangeValue: this.opacityRangeConfig.value,
       hueRangeValue: this.hueRangeConfig.value
     };
+
+    this.closeTooltip = this.closeTooltip.bind(this);
+  }
+
+  shouldComponentUpdate(nextProps) {
```
nerik: `shouldComponentUpdate` is not the right place for that. If your `shouldComponentUpdate` implementation always returns the same value, it means you should not use `shouldComponentUpdate`.
Use another React comp lifecycle hook, like `componentWillUpdate()`, `componentDidUpdate()` or even `render` depending on the use case.

app/src/components/Map/LayerBlendingOptionsTooltip.jsx
```diff
@@ -46,6 +47,23 @@ class LayerOptionsTooltip extends Component {
...
+    if (nextProps.visible !== this.props.visible) {
+      if (nextProps.visible) window.addEventListener('click', this.closeTooltip);
+      else window.removeEventListener('click', this.closeTooltip);
```
nerik: I actually don't think you need to add/remove that event listener at each component update. This could be done at component mount/unmount.
Your implementation has a lot of potential for memory leaks (event listeners never cleanup)

nerik:
Apart from those comments: could you please check why the icon is not highlighted in white anymore when opening the tooltip?
Thanks !

app/src/components/Map/LayerBlendingOptionsTooltip.jsx
sorodrigo: You need to add/remove it cause if you leave it the tooltip will open/close on any action due to the fact that the component is always mounted.

app/src/components/Map/LayerBlendingOptionsTooltip.jsx
sorodrigo: really smooth Rodrigo... Thanks I'll change it.

sorodrigo:
@nerik ready!

nerik:
Yeah.

PR #539, opened 2017-02-16
prevents zoom on large spanned custom layers

sorodrigo:
This PR fixes an issue where large layers (global ones specially) were zoomed out and map went gray.

### DIFF
--- a/app/src/components/Layers/CustomLayerWrapper.js
+++ b/app/src/components/Layers/CustomLayerWrapper.js
@@ -1,10 +1,15 @@
+import { MAX_AUTO_ZOOM_LONGITUDE_SPAN } from 'constants';
+
 export default class CustomLayerWrapper {
   constructor(map, url) {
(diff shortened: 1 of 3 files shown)


sorodrigo:
For testing:
- Global http://storage.googleapis.com/world-fishing-827/scratch/enrique/oceans.kml
- Local http://googlemaps.github.io/js-v2-samples/ggeoxml/cta.kml

nerik:
Awesome!
One small comment: you are never cleaning up the event listener set up in the constructor of `CustomLayerWrapper`. Maybe now is the good time to actually implement a remove mechanism for custom layers, because I totally forgot 😝 
See https://github.com/Vizzuality/GlobalFishingWatch/blob/develop/app/src/components/Layers/MapLayers.jsx#L275
Basically call `hide`, clean up the event, done

nerik:
Thanks !

PR #541, opened 2017-02-17
Mobile timebar

nerik:


nerik:
screen shot 2017-02-17 at 15 45 51

PR #542, opened 2017-02-17
Immediate map update on inner brush movement; stop playback on timeba…

tiagojsag:
…r interactions


nerik:
Looks like the max inner range bounding got lost in the way.
Also, nice to have: throttle when resizing (not moving), because this can get quite slow

PR #544, opened 2017-02-17
Feature: custom layer ACL

sorodrigo:
This PR adds ACL to allow/prevent upload of custom layers.


app/src/components/Map/LayerManagementModal.jsx
```diff
@@ -3,7 +3,7 @@ import classnames from 'classnames';
 import LayerLibrary from 'containers/Map/LayerLibrary';
 import CustomLayer from 'containers/Map/CustomLayer';
 
-import LayerMaganementModalStyles from 'styles/components/map/c-layer-management-modal.scss';
+import LayerManagementModalStyles from 'styles/components/map/c-layer-management-modal.scss';
```
tiagojsag: nice catch ;)

tiagojsag:
Looks great :) I did a minor refactoring on the HTML code: I usually prefer to have as little logic as possible in the middle of the HTML code, as it's usually very easy to miss it.

PR #546, opened 2017-02-20
Remove site code

tiagojsag:
Known issues:
- review what npm deps we can remove


nerik:
I left you a TODO but otherwise I think it's working (didn't have the time to test log in/out though)

PR #547, opened 2017-02-20
Vessel metrics based on "real" data

tiagojsag:


app/src/components/Map/ControlPanel.jsx
```diff
@@ -32,30 +32,52 @@ class ControlPanel extends Component {
...
+    const min = this.props.timelineInnerExtent[0].getTime();
+    const max = this.props.timelineInnerExtent[1].getTime();
+
+    if (this.props.chartData.length === 0) {
+      return '-';
+    }
+
+    let result = this.props.chartData.reduce((acc, elem) => {
+      if (elem.date >= min && elem.date <= max) {
+        return acc + elem.value;
```
nerik: This belongs to a reducer. For instance listen to SET_INNER_TIMELINE_DATES in the timebar reducer.

app/src/components/Map/ControlPanel.jsx
```diff
@@ -32,30 +32,52 @@ class ControlPanel extends Component {
...
+      }
+      return acc;
+    }, 0);
+
+    result = `${Math.round(result)}`;
+    for (let i = result.length - 3; i > 0; i -= 3) {
```
nerik: Please replace by something more concise:
```
result.toLocaleString()
```
Performance is not good on 1000s of elemements, but in that case with one element the overhead is neglectible.

app/src/components/Map/ControlPanel.jsx
```diff
@@ -139,15 +161,11 @@ class ControlPanel extends Component {
...
-            <span className={controlPanelStyle['counter-description']}>vessels</span>
-            <span className={controlPanelStyle.total}>224,654</span>
-          </div>
-          <div className={controlPanelStyle['activity-display']}>
-            <span className={controlPanelStyle['counter-description']}>activity</span>
-            <span className={controlPanelStyle.total}>224,654</span>
+      <div className={controlPanelStyle['resume-display']} >
+        <div className={controlPanelStyle['categories-display']} >
+          <div className={controlPanelStyle['vessel-display']} >
+            <span className={controlPanelStyle['counter-description']} >vessels</span>
```
nerik: Careful. You are not showing the number of vessels at all, but rather the fishing activity for a given time range, expressed in a unitless value.

app/src/components/Map/ControlPanel.jsx
tiagojsag: To me, this is computed view data, not real app state. Adding this to the state container is an implicit state duplication, which is what we're trying to avoid.

app/src/components/Map/ControlPanel.jsx
tiagojsag: I updated the json file to have large int values - something like "number of vessels" or "number of fishing events". Anyway, this is not yet defined, so let's not worry too much about it until we know exactly what we'll have here

app/src/components/Map/ControlPanel.jsx
tiagojsag: changed

app/src/components/Map/ControlPanel.jsx
sorodrigo: Might be interesting: https://github.com/reactjs/reselect

Btw vue.js has this out of the box...

PR #549, opened 2017-02-28
Feature/save current vessel on workspace

nerik:


nerik:
Support for legacy workspace works in theory, but our map breaks with the 801 tileset, which makes it a bit hard to test.

PR #550, opened 2017-03-01
Playback auto restart

tiagojsag:
@nerik doesn't feel like the DRYest code ever, but I can't seem to improve it. thoughts?


nerik:
I did a slightly less DRY solution, but more importantly it relies more on the app state to do the job.

tiagojsag:
@nerik I changed the actual feature to my original functionality - return to start will only happen if you press play when you are at the end. Let's discuss this later if you want

PR #554, opened 2017-03-02
fixes bizarre tracks rendering if out of viewport

sorodrigo:


nerik:
Last commit should fix issues with both tracks out of viewport and tracks crossing the antimeridian @sorodrigo

PR #558, opened 2017-03-08
Fix: transforms dates to utc

sorodrigo:
This PR fixes an issue that ocurred when loading timebar json files from timezones that were still in 2011 at `2012-01-01 UTC` time.


tiagojsag:
https://www.w3schools.com/jsref/jsref_getutcfullyear.asp ?

sorodrigo:
@tiagojsag well… this is embarrassing. Thanks, it's fixed now!

PR #565, opened 2017-03-20
Fixes broken reporting id field lookup.

andres-arana:
Seems like cartodb doesn't support uppercase identifiers in certain
fields. When we lookup the `reportingId` field, cartodb always returns
`reportingid` (lowercase i), and thus the app never finds that field,
defaulting to using the `cartodb_id` field instead on report requests.

This commit changes the field lookup to be `reporting_id` instead,
which doesn't have this problem and it also makes it more consistent
with the rest of the identifiers we have on carto.


nerik:
Hey @andres-arana, thanks for the fix. I've added a fallback to `reportingId` just in case. Not very elegant but we should be covered for all cases.

PR #569, opened 2017-04-04
Feature/highlight track

nerik:
Note:
Negative seriesgroup are still ignored (they don't match a vessel but a vessel cluster) - check with Skytruth if this is the intended behaviour for this feature.


nerik:
fishing-highlight-0

fishing-highlight

tiagojsag:
The highlighted vessels are hard to see in areas of the map with lots of fishing effort. Can we improve this?

nerik:
fishing-highlight-dim

PR #578, opened 2017-04-18
Fix/tracks gaps

nerik:
- Fixes issue with gaps in tracks rendering (https://basecamp.com/1756858/projects/12049853/todos/303904497)
- changed default track color to be the same pink as heatmap highlight (addresses https://basecamp.com/1756858/projects/12049853/todos/304548972)


nerik:
screen shot 2017-04-18 at 18 37 45

tiagojsag:
why pink?

PR #581, opened 2017-04-20
Add custom layers before heatmap layer

tiagojsag:
Add custom layers before heatmap layer


app/src/reducers/layers.js
```diff
@@ -1,15 +1,15 @@
 import _ from 'lodash';
 import {
-  SET_LAYERS,
```
tiagojsag: dunno what my IDE did here, most likely sort these alpahbetically.

nerik:
screen shot 2017-04-20 at 13 16 07

Doesn't seem to be working for me, or I am missing something. It might be a bit more complicated than fiddling with the layers state, as a new layer is always added on top https://github.com/Vizzuality/GlobalFishingWatch/blob/develop/app/src/components/Layers/MapLayers.jsx#L246

nerik:
Overall we haven't been dealing with layer order properly. It's on a last arrived, last on top basis, but I think we should use Google Maps panes or something.

nerik:
(I can have a look next week if you prefer)

PR #582, opened 2017-04-20
Use layer id on search

tiagojsag:


nerik:
At some point we'll need to think about search across more layers than just the vessels one, but this is perfect for now.

PR #585, opened 2017-04-21
decouple layer id from tileset id

tiagojsag:
We had an implicit requirement that a fishing layer's id in the workspace had to equal the tileset id. This removes that restrictions, which should help on some scenarios. It also puts us 1 step closer to multi-layer vessel searches.

overall, I feel like the map.tilesetUrl and map.tilesetId states should be deprecated.

### DIFF
--- a/app/src/actions/layers.js
+++ b/app/src/actions/layers.js
@@ -102,9 +102,6 @@ export function initLayers(workspaceLayers, libraryLayers) {
           description: libraryLayer.description || matchedWorkspaceLayer.description,
           reportId: libraryLayer.reportId
         });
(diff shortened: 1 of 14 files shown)


app/src/actions/layers.js
```diff
@@ -103,6 +102,9 @@ export function initLayers(workspaceLayers, libraryLayers) {
           description: libraryLayer.description || matchedWorkspaceLayer.description,
           reportId: libraryLayer.reportId
         });
+        if (matchedWorkspaceLayer.type === LAYER_TYPES.Heatmap) {
```
tiagojsag: Need to remove this and move it elsewhere

PR #586, opened 2017-04-24
Feature/heatmap perf

nerik:
These changes make panning and zooming much more responsive.


tiagojsag:
lots of console.log and commented code. feels itchy to me, but if you want to keep them, fine by me. you decide

PR #589, opened 2017-04-26
show fishing activity on tracks

nerik:
screen shot 2017-04-26 at 18 39 21

I just reproduced the previous version's behaviour for now.


nerik:
@tiagojsag checked with JC, this still needs to be refined a bit, but it should adress teh concern Skyruth had with rendering of track dots.

PR #591, opened 2017-04-28
fixed frozen heatmap while zooming

nerik:


.gitignore
```diff
@@ -16,3 +16,6 @@
 
 # Compiled files
 /dist
+
```
tiagojsag: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

nerik:
Nice catch! Should be fixed.

.gitignore
tiagojsag: can you remove this from here, please? I really want to avoid IDE/system specific stuff on project's gitignores

PR #595, opened 2017-04-28
Fix/colors

nerik:
fishing-saturation-tracks
fishing-saturation-tracks2


tiagojsag:
image
can we make the dot size depend on something? when we have a lot of clustered dots, it become unreadable

tiagojsag:
also, the far right edge of the hue slider bar has a weird effect on the colors

nerik:
Issue with right edge of hue slider is fixed.
Re dot size: size only depends on zoom level. I've changed the calculation slightly so that when zooming in, dots appear bigger more slowly, which should fix the effect you showed at medium zoom levels.

PR #597, opened 2017-05-02
fixed time extents and heatmap lookup with legacy workspaces

nerik:


app/src/actions/workspace.js
```diff
@@ -215,7 +215,7 @@ function processLegacyWorkspace(data, dispatch) {
     endOuterDate = new Date(serializedStartDate[1]);
     startOuterDate = new Date(endOuterDate.getTime() - workspace.state.timeExtent);
     startInnerDate = startOuterDate;
-    endInnerDate = new Date(startInnerDate.getTime() + Math.min(workspace.state.timeExtent, 15778476000));
+    endInnerDate = new Date(startInnerDate.getTime() + Math.min(workspace.state.timeExtent / 2, 15778476000));
```
tiagojsag: ??

I dont understand what this fixes. Seems to only inject randomness. could you explain?

.gitignore
```diff
@@ -16,3 +16,5 @@
 
 # Compiled files
 /dist
+
+private-indonesia-workspace-v1.json
```
tiagojsag: this probably should not be here, right?

nerik: Sorry it's for my own convenience, I'll put this in my global gitignore

app/src/actions/workspace.js
nerik: Well if I understood correctly, when loading legacy workspaces, you'll always end up with inner range === outer range (since timeExtent is used for both). 
This allows the UI to be loaded with a state that makes it more obvious that you can move the inner brush (since it doesn't take up the whole width). 
But yeah, I agree that half of it is kind of arbitrary, let me know what you think works best.

PR #603, opened 2017-05-05
Fix map center loading

tiagojsag:
because we were using an array to map the map center in props, differences were never detected. So loading a map from workspace would result in the center being 0,0

this should fix the issue


nerik:
👍

PR #608, opened 2017-05-09
Fix/improve workspace export and import

tiagojsag:


nerik:
It looks like pinned layers are now set to visible = false by default, which is not the intended behaviour.
- select track on map
- pin it
- close vessel details
- track does not show up anymore, and toggle is off in pinned tracks list

nerik:
Otherwise workspace save/import now works correctly with pinned tracks 👍 
The only thing is that vessels order is not preserved, but I guess it's less important.

tiagojsag:
that commit should fix it

As for the order, we should fix it, but another day

PR #609, opened 2017-05-10
pinned vessel title fixes

nerik:
- fallback title when vesselname is not present (https://basecamp.com/1756858/projects/12049853/todos/307159441)
- fixed a bug where typing a custom would be very slow when pinned track was visible

### DIFF
--- a/app/src/reducers/vesselInfo.js
+++ b/app/src/reducers/vesselInfo.js
@@ -50,8 +50,13 @@ export default function (state = initialState, action) {
     case SET_VESSEL_DETAILS: {
       const vesselIndex = state.vessels.findIndex(vessel => vessel.seriesgroup === action.payload.seriesgroup);
       const currentVessel = state.vessels[vesselIndex];
+      const defaultTitle =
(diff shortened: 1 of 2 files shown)


app/src/reducers/vesselInfo.js
```diff
@@ -166,7 +171,7 @@ export default function (state = initialState, action) {
     }
     case SET_PINNED_VESSEL_TITLE: {
       const vesselIndex = state.vessels.findIndex(vessel => vessel.seriesgroup === action.payload.seriesgroup);
-      const newVessel = _.cloneDeep(state.vessels[vesselIndex]);
+      const newVessel = Object.assign({}, state.vessels[vesselIndex]);
```
tiagojsag: what's the difference here?

nerik: Object.assign makes a shallow copy of the object while cloneDeep makes a copy in memory of the whole objects (all nested objects). This caused typing a custom title to be very slow because it was copying the whole tracks array that is attached to the vessel.

PR #613, opened 2017-05-10
Fallback canvas renderer

nerik:
screen shot 2017-05-10 at 18 49 42


nerik:
screen shot 2017-05-10 at 19 11 39

PR #619, opened 2017-05-16
Dismissable legacy banner

nerik:
screen shot 2017-05-16 at 12 27 43


tiagojsag:
Use https://github.com/Vizzuality/GlobalFishingWatch/blob/develop/public/literals.json instead of an env var.

tiagojsag:
If you close the banner, this happens: (chrome macOS)

image

And if the sidebar grows, this happens:

image

nerik:
Thanks, will fix.

PR #620, opened 2017-05-16
Fix/bundle weight

nerik:
Before:

screen shot 2017-05-16 at 16 41 25

After:

screen shot 2017-05-16 at 17 53 37

Also did a bit of cleanup, bumping dependency versions, etc


nerik:
(merge after release)

app/src/components/Map/FilterPanel.jsx
```diff
@@ -66,7 +67,7 @@ class FilterPanel extends Component {
   }
 
   updateFilters(filter, index) {
-    const updatedFilters = _.cloneDeep(this.props.flags);
+    const updatedFilters = cloneDeep(this.props.flags);
```
sorodrigo: I'm not sure this is needed. I think that you can clone only the part of the tree that needs to be updated, and avoid re-rendering the rest. A shallow clone might do the trick.
That said, this might not be the case so we would need to do some testing.
[More info]
https://twitter.com/dan_abramov/status/688087202312491008?lang=es

app/src/components/Map/FilterPanel.jsx
```diff
@@ -87,7 +88,7 @@ class FilterPanel extends Component {
   }
 
   removeFilter(index) {
-    const filters = _.cloneDeep(this.props.flags);
+    const filters = cloneDeep(this.props.flags);
```
sorodrigo: same as above

app/src/components/Map/LayerBlendingOptionsTooltip.jsx
```diff
@@ -25,16 +27,16 @@ class LayerBlendingOptionsTooltip extends Component {
   constructor(props) {
     super(props);
 
-    this.opacityRangeConfig = _.cloneDeep(INPUT_RANGE_DEFAULT_CONFIG);
-    _.assign(this.opacityRangeConfig, {
+    this.opacityRangeConfig = cloneDeep(INPUT_RANGE_DEFAULT_CONFIG);
```
sorodrigo: same as above

app/src/components/Map/LayerBlendingOptionsTooltip.jsx
```diff
@@ -25,16 +27,16 @@ class LayerBlendingOptionsTooltip extends Component {
...
+    assign(this.opacityRangeConfig, {
       minValue: 10,
       maxValue: 100,
       step: 1,
       value: this.props.opacityValue * 100
     });
 
-    this.hueRangeConfig = _.cloneDeep(INPUT_RANGE_DEFAULT_CONFIG);
-    _.assign(this.hueRangeConfig, {
+    this.hueRangeConfig = cloneDeep(INPUT_RANGE_DEFAULT_CONFIG);
```
sorodrigo: same as above

app/src/reducers/layers.js
```diff
@@ -14,7 +15,7 @@ import {
 import { LAYER_TYPES } from 'constants';
 
 const getUpdatedLayers = (state, action, changedLayerCallback) => {
-  const layers = _.cloneDeep(state.workspaceLayers);
+  const layers = cloneDeep(state.workspaceLayers);
```
sorodrigo: same as above

app/src/reducers/layers.js
```diff
@@ -100,7 +101,7 @@ export default function (state = initialState, action) {
       });
 
       if (newState.layerPanelEditMode === false) {
-        newState.workspaceLayers = _.cloneDeep(state.workspaceLayers);
+        newState.workspaceLayers = cloneDeep(state.workspaceLayers);
 
 
         newState.workspaceLayers.forEach((layer) => {
```
sorodrigo: same as above. Perhaps here, we could use `map` instead of `forEach` and return new objects where needed,

nerik:
@sorodrigo I agree 100% with you about cloneDeep, which we used way too liberally throughout the project. In a few places where mutating data is not actually problem, others where it's just a lazy replacement for other ways to prevent mutating (concat, Object.assign, etc).
I wanted to keep this PR focused on cleaning up dependencies though (without touching any logic), and I believe this would be better done in a separate PR.

tiagojsag:
I see replacing those cloneDeep with something else as one of those changes that seem harmless but actually mess up a few things. I fully support it, but only after the release, and with proper testing

tiagojsag:
@nerik do you have any metrics on how much we saved, in terms of data transfered, with these changes?

sorodrigo:
Ok I agree we can address the `cloneDeep` in the future. It's good to go then!

nerik:
@tiagojsag overall I think this PR should be merged after the release.
We went from a 13.8mb bundle to 10.6mb. 
For more, replacing React by preact would be rather easy to do. 
Bit less easy, replace moment with something more lightweight, like https://date-fns.

**Note: can you please make sure with Skytruth that their env var NODE_ENV is set to production ? This activates minification, going from 10.6mb to 1.5mb**

tiagojsag:
I'm putting WIP just so we know we cant merge now (I tend to get trigger happy sometimes)

nerik:
Let me know when we can start testing this on staging please, I'd very much like this to not get stale.
Thanks :)

PR #623, opened 2017-05-17
fixed loading of AIS/VMS workspace

nerik:
- there was an issue with the workspace, since both heatmap layers had the same layer id.
- I added handling of mismatch between columnIds available on tileset headers vs columns that are actually available in every tile, which seemed to happen at some point.


nerik:
Tested tracks rendering with both layers, it's working fine.

PR #625, opened 2017-05-23
Add support for dynamic vessel info

tiagojsag:
Until now, vessel details used a hardcoded strucutre. since we now have multiple layers, each with different vessel info details, we need to rely on each vessel layer header to tell us which field and what type it is.

### DIFF
--- a/app/src/components/Map/VesselInfoPanel.jsx
+++ b/app/src/components/Map/VesselInfoPanel.jsx
@@ -31,7 +31,10 @@ class VesselInfoPanel extends Component {
   }
 
   render() {
-    const vesselInfo = this.props.vessels.find(vessel => vessel.shownInInfoPanel === true);
+    // const vesselInfo = this.props.vessels.find(vessel => vessel.shownInInfoPanel === true);
+    // instead directly use:
+    const vesselInfo = this.props.currentlyShownVessel.vessel;
+
     const status = this.props.infoPanelStatus;
 
     if (status === null && vesselInfo === undefined) {
@@ -49,7 +52,10 @@ class VesselInfoPanel extends Component {
     } else if (this.props.userPermissions !== null && this.props.userPermissions.indexOf('seeVesselBasicInfo') === -1) {
       return null;
     } else if (vesselInfo !== undefined) {
-      const currentLayer = this.props.layers.find(layer => layer.tilesetId === vesselInfo.tilesetId);
+      // const currentLayer = this.props.layers.find(layer => layer.tilesetId === vesselInfo.tilesetId);
+      // instead directly use:
+      const currentLayer = this.props.currentlyShownVessel.layer;
+      // I haven't checked but we could probably go further directly collecting vesselFields in the action, instead of sending the whole layer
       let layerFields;
       if (currentLayer !== undefined && currentLayer.header !== undefined && currentLayer.header.vesselFields !== undefined) {
         layerFields = currentLayer.header.vesselFields;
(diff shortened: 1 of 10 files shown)


nerik:
Haven't tested the feature yet, are you using workspace_vms.json ?

Looking at the code, a few comments: 
- let's try to start enforcing the rule we set for ourselves: at least a short description and a bit of context in the PR description, a gif of the feature in action even better.
- it seems to me that we are messing way to much with the state in the VesselInfo view (all vessels array and all layers array). I've mocked a few suggested changes here https://github.com/Vizzuality/GlobalFishingWatch/pull/625/commits/13655a964a1847832ce64d74e78cd1906b338d1f
it's not functional, but should give you an idea. LMK what you think.
- is there a particular reason you changed the hue to a color in the workspace? Shouldn't we use hue by default (otherwise the actual vessel hue will fallback to 0 I think)

tiagojsag:
1- done
2- commented in your commit, but I agree with you
3- I copied it over from somewhere else. we actually convert it to a hue. I can revert that, if needed

nerik:
Thanks 👍 
Since the vessel's `tilesetId` is used for the layer lookup, it should be a problem, or am I missing something?

app/src/components/Map/VesselInfoPanel.jsx
```diff
@@ -31,13 +31,10 @@ class VesselInfoPanel extends Component {
...
   render() {
-    // const vesselInfo = this.props.vessels.find(vessel => vessel.shownInInfoPanel === true);
-    // instead directly use:
-    const vesselInfo = this.props.currentlyShownVessel.vessel;
-
+    const vesselInfo = this.props.currentlyShownVessel;
     const status = this.props.infoPanelStatus;
 
-    if (status === null && vesselInfo === undefined) {
+    if (!status && !vesselInfo) {
```
nerik: I say never, ever rely on truthy/falsy values in JS, but it seems I'm the only one 😝 
We should remember that :
- 0, null and undefined can mean very different things.
- explicit equality checks (=== undefined) also serve as type hints in modern JS engines (better perfs)
- explicit equality checks are just easier to read

app/src/containers/Map/VesselInfoPanel.js
```diff
@@ -3,13 +3,20 @@ import VesselInfoPanel from 'components/Map/VesselInfoPanel';
...
 import { login } from 'actions/user';
 
-const mapStateToProps = state => ({
-  // layers: state.layers.workspaceLayers,
-  // vessels: state.vesselInfo.vessels,
-  currentlyShownVessel: state.vesselInfo.currentlyShownVessel,
-  infoPanelStatus: state.vesselInfo.infoPanelStatus,
-  userPermissions: state.user.userPermissions
-});
+const mapStateToProps = (state) => {
```
nerik: 👍

app/src/constants.js
```diff
@@ -122,6 +122,10 @@ export const REPORT_STATUS = {
   sent: 'sent'
 };
 
+export const VESSEL_INFO_HIDDEN = 1;
+export const VESSEL_INFO_LOADING = 2;
+export const VESSEL_INFO_LOADED = 3;
```
nerik: In that kind of things I tend to prefer an "enumerable" style, but probably a matter of taste:
```
export const VESSEL_INFO_STATUS = {
  HIDDEN: 1
  ...
} 
...
if (status === VESSEL_INFO_STATUS.HIDDEN) 
...

```

tiagojsag: :+1:

PR #627, opened 2017-05-29
Only filter by flag on layers that have category field

tiagojsag:
This makes sure that, when filtering by flag on the heatmap rendering, we check for the existence of the "category" data before accessing it.


nerik:
👍

PR #628, opened 2017-05-30
added reportable layers warning in report panel

nerik:
TODO: 
- <s>send actual value for `reportableHeatmapLayers`</s>
- <s>toggle message off if there is only 1 heatmap layer and it's reportable?</s>


nerik:
reportable

PR #632, opened 2017-07-13
Optimize tile loading

nerik:
(see below👇 )


nerik:
### Why
This is a proof of concept whose goal is to enhance tile loading, meaning:
- reduce the time needed from tile request to vessels display;
- reduce UI 'freezing' when decoding tiles.
Both happening at initial page load, map zooming and map panning.

### Benchmark
Load 24 tiles at zoom level 3 (typical initial loading on a laptop screen):

Before: 
total processing time: ~ 900 ms
total size: ~ 6 MB of data

After: 
total processing time: ~ 350 ms (about 3 times faster)
total size: ~ 2 MB of data (about 3 times less heavy)

### How ?

The faster response times are due to "baking" computation previously done on the CPU in the tiles, offline. World coordinates (projected lat and lon), as well as heatmap brush opacity and radius (both dependant on zoom level) are directly stored inside the tiles. 

Tiles are made much smaller (while retaining more data) by using Google Protobuf storage format (see below for the implementation used).

Additionally, all values are stored as integers, as protobuf is much less efficient at storing and decoding floats.


### PoC implementation

This PR is merely a PoC, meaning that the various libs and tools used can probably be substituted.

_Test data:_
1. a big GeoJSON file (about 40mb of raw data) is prepared directly in the client (it contains the data of 24 currently visible tiles)
https://github.com/Vizzuality/GlobalFishingWatch/pull/632/files#diff-0eeda153084002b1f9f50e1f8270d43bR23

_Offline:_
2. The GeoJSON is converted to an **mbtiles** file (sqlite3 wraper), using **tippecanoe**
https://github.com/nerik/fishing-vector/blob/master/package.json#L9
_Note: in this we are using tippecanoe in a very simplistic way, but one of its strength is actually to generate aggregated geometries at lower zoom levels._

_Server-side:_
3. Tiles are served from the mbtiles file as .pbf tiles (**protobuf**) by a very simple **Tilelive**/Express server
https://github.com/nerik/fishing-vector/blob/master/server.js
 
_Client-side:_

4. Tiles are loaded, then pbf data is decoded
https://github.com/Vizzuality/GlobalFishingWatch/pull/632/files#diff-cfce8bd2e1e0bfe899c82e36d1f7a67fR26

5. pbf are converted to <a href="https://github.com/mapbox/vector-tile-spec">**Mapbox vector tiles**</a> features 
https://github.com/Vizzuality/GlobalFishingWatch/pull/632/files#diff-cfce8bd2e1e0bfe899c82e36d1f7a67fR28
https://github.com/Vizzuality/GlobalFishingWatch/pull/632/files#diff-4d093f9b78b2247a5ed425b89975b108R168

6. Integers are converted back to floats (with a given multiplier)
https://github.com/Vizzuality/GlobalFishingWatch/pull/632/files#diff-4d093f9b78b2247a5ed425b89975b108R173

### Future-proof

The benefit of using Mapbox vector tiles now, is that it opens the possibility of using the super smooth Mapbox GL client, on the short-medium term. Here's a test with the same tiles:

fish-vector

It's all the more interesting considering Mapbox is currently developing a heatmap mode:
https://blog.mapbox.com/sneak-peek-at-heatmaps-in-mapbox-gl-73b41d4b16ae

nerik:
On track rendering performance, see 
https://github.com/Vizzuality/GlobalFishingWatch/pull/633

PR #633, opened 2017-07-17
Tracks perf

nerik:
Trying to improve track performance by reducing the number of points. 
The number of points can be divided by 3 without any visible impact, and divided by 10 with a barely noticeable impact.
As can be seen below, the algo should preserve fishing points, and currently it doesn't. But, I'm not sure I should keep on improving this, because ideally, this should be done on the backend anyways (ie it doesn't make sense to download the points coordinates while 90% of them would eventually be discarded)

original: 30 000 points

track-simplify-original

tolerance 0.01: 10 000 points

track-simplify-0 01-3

tolerance 0.1 : 3 000 points

track-simplify-0 1-10

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+http://localhost:3000/?workspace=udw-v2-6a974c26-35e6-46e2-9064-15023ca8eed6
+
 # Global Fishing Watch
 
 This repository will host the Global Fishing Watch client application developed by [vizzuality](http://www.vizzuality.com/)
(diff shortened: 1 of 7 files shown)


nerik:
Probably not needed for now with https://github.com/Vizzuality/GlobalFishingWatch/pull/637
Have a look later. 
Need to fork simplify-js to 
- prevent it from skipping fishing points (only using simple radial dist should be enough)
- make it work with array vectors

PR #634, opened 2017-07-17
switched to preact

nerik:
Shaves about 1.5 mb of 10.6 mb of the unminifed bundle. Also might improve performance a little bit. Not a revolution, but a quick win anyways :)


nerik:
LGTM no? @tiagojsag

PR #635, opened 2017-07-18
removed seriesUid

nerik:
Removed a part of the vessels frames that was not needed


tiagojsag:
http://localhost:3000/?workspace=udw-v2-7f3f4504-c9a6-443d-ae67-10de8365d2e4

On the workspace above, if you click on some of the vessels on the top part of the map, it hides those tracks - instead of adding them, as they are pinned

image

image

tiagojsag:
on second thought, they seem to be the same boat, just different routes. what did we have in mind for these cases?

nerik:
@tiagojsag it's an edge case we need to tackle indeed. Maybe add this to the upcoming feature list.

PR #638, opened 2017-07-25
added North Star basemap

nerik:
fishing-northstar

fishing-northstar-2


tiagojsag:
Let's run this by Enrique and/or Paul, and if they like it, we need to move it to their mapbox account

nerik:
Note that we are already using other Mapbox basemaps (with Dani's token)
The question is : do we make this a default?

danielcasius:
It looks amazing. Great job, @nerik !

tiagojsag:
The other basemaps point to enrique's account. If we are indeed using Dani's account, we should not

nerik:
Ah my bad, it's not the same token

tiagojsag:
@enriquetuya this is part of the new design iteration, but since it's already implemented, we can deploy it as soon as we want - we just need to export to you Mapbox account and change the URL. Let me know how you wish to proceed

tiagojsag:
https://basecamp.com/1756858/projects/12049853/todos/319956461

PR #703, opened 2017-08-04
Use CSS modules properly

nerik:
This is following a conversation with @vectorsize and @j8seangel about CSS modules.
We have been using them in a not very practical way on this project from the start.

This PR: 
-  drops the BEM-like conventions (`.some-component, .-modifier`) for camel case (`.someComponent, ._modifier`). 
This allows using css modules in React components in a much saner way. Before: 
```
        <div
          className={classnames(
            searchPanelStyles['results-container'],
            { [`${searchPanelStyles['-open']}`]: this.props.searchResultsOpen }
          )}
        >
```

After:
```
        <div
          className={classnames(
            searchPanelStyles.resultsContainer,
            { [searchPanelStyles._open]: this.props.searchResultsOpen }
          )}
        >
```

- limits nesting in SCSS files to 1 level (was 4 before). This is not enforced across all files yet, but the reasoning is the following: 
Less nesting means :
  - a much more readable and maintainable SCSS codebase (ie not having to find that closing bracket 500 lines below).
  - less heavy CSS output
The benefit of CSS modules is that it will by definition namespace a SCSS file / remove the global namespace - so no more nesting is needed.

- cleans up, removes unused code, adds sanity


nerik:
I've also tried to harmonize the way we name CSS imports within React components, should be:
```
import ComponentStyles from 'styles/components/component.scss';
```
Let's try to avoid things such as 
```
import css from ...
```

PR #734, opened 2017-08-08
Feature/serialize areas

Bluesmile82:
image

Save the areas to the workspace and add color to the areas model. Includes the new colorPicker and changing the polygon color during editing. All the changes are saved in a temporary area in the store called editingArea.

- [x] Save areas to workspace
- [x] Add color to areas

### DIFF
--- a/app/src/components/Map/AreasPanel.jsx
+++ b/app/src/components/Map/AreasPanel.jsx
@@ -1,7 +1,9 @@
 import React, { Component } from 'preact';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import ColorPicker from 'components/Shared/ColorPicker';
 import areasPanelStyles from 'styles/components/map/areas-panel.scss';
+import controlPanelStyles from 'styles/components/control_panel.scss';
 import buttonStyles from 'styles/components/map/button.scss';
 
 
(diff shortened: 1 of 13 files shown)


app/src/containers/Map/DrawingManager.js
```diff
@@ -1,14 +1,15 @@
 import { connect } from 'react-redux';
 import DrawingManager from 'components/Map/DrawingManager';
-import { saveCoordinates } from 'actions/areas';
+import { saveEditingArea } from 'actions/areas';
 
 const mapStateToProps = state => ({
-  map: state.map.googleMaps
+  map: state.map.googleMaps,
+  polygonColor: state.areas.editingArea.color,
```
tiagojsag: linter is complaining here. use the git hooks to validate before committing - more info Monday

app/src/components/Map/AreasPanel.jsx
```diff
@@ -38,10 +43,15 @@ class AreasPanel extends Component {
...
+  }
+
+  onColorChange(color) {
+    this.props.saveEditingArea({ color });
   }
 
   render() {
+    const { name, color } = this.props.editingArea;
     return (
       (this.props.drawing ?
```
tiagojsag: Rule of thumb: only use the trenary operator if each part of it takes 1 line at most. 

When you have more than 1 line, the trenary operator is easy to miss. In those situations, use a traditional if/else clause.

PR #746, opened 2017-08-14
Fix menu style

Bluesmile82:
Fix new menu styling:
image

- [x] Vessel info should be always present
- [x] Fix search list overflow
- [x] Line for the first link in the main menu
- [x] Layer hue and opacity tooltip is clipped
- [x] Fix mobile display
- [x] Hide area submenu
- [x] Animation for menu
- [x] Configure autoprefixer


tiagojsag:
- [ ] Try and make the animation smoother and faster
- [ ] Top logo align issue is back
- [ ] Make the menu titles also clickable to navigate back to the main menu

PR #751, opened 2017-08-17
Prevent unecessary cloning of vessel tracks on vessel edits

tiagojsag:
We are doing deep cloning way more often than needed. For example, pinning/unpinning a vessel can take from 0.5 to a couple seconds because of it.

Silly approach (implemented here): replace all cloneDeep calls with plain clone calls. Risky but it seems to work, as I think we never really compare the track array itself, only the vessel object

Pro approach (maybe for later): keep tracks separated from vessel details in the reducer, and use an id for cross-access. 

thoughts?


nerik:
Very nice, this has been bugging me for a while. The only remark I have: instead of using `_.clone` I'd rather see the native `Object.assign`.

PR #752, opened 2017-08-17
Timebar changes

Bluesmile82:
image

- [x] Reposition Datepickers and change styles
- [x] Change play button

### DIFF
--- a/app/src/timebar/components/DatePicker.jsx
+++ b/app/src/timebar/components/DatePicker.jsx
@@ -25,7 +25,7 @@ class DatePicker extends Component {
         </div>
         <ReactDatePicker
           fixedHeight
-          calendarClassName={classnames(['calendar', { rightCalendar: this.props.label === 'end' }])}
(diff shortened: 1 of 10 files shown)


app/src/timebar/components/DatePicker.jsx
```diff
@@ -24,6 +25,7 @@ class DatePicker extends Component {
         </div>
         <ReactDatePicker
           fixedHeight
+          calendarClassName={classnames(['calendar', { endDatePicker: this.props.label === 'end' }])}
```
tiagojsag: Us-in-6-months will change the label from "end" to "end date" or "finish date" or something else, and we'll have a CSS regression. Best to add a new, more explicit prop, like "calendarClassName"

Bluesmile82: But this should be only applied to the calendar on the right of the screen. Maybe rightCalendar could be more explicit about the styling

tiagojsag: ok, but then we should add a new prop to handle that.

PR #753, opened 2017-08-17
Feature/basemap inside layers menu

Bluesmile82:
This branch depends on fix/menu-style

image

- [x] Include basemap choice in layers menu and style
- [x] Remove basemap submenu


app/src/rightControlPanel/components/ControlPanel.jsx
```diff
@@ -90,6 +93,7 @@ class ControlPanel extends Component {
       this.props.setRecentlyCreated(false);
       this.props.setDrawingMode(false);
     }
+    this.setState({ showSubmenu: false });
```
tiagojsag: this seems like duplicate state from this.props.submenu. can it be removed?

app/src/rightControlPanel/components/ControlPanel.jsx
```diff
@@ -125,39 +129,49 @@ class ControlPanel extends Component {
...
+      </div>
     );
 
     const layerSubmenu = (
-      <SubMenu title="Layers" icon={this.renderIcon('layers')} onBack={this.onBack} >
+      <SubMenu
+        title="Layers"
+        icon={this.renderIcon('layers')}
+        onBack={this.onBack}
+        contentBefore={<BasemapPanel />}
```
tiagojsag: I don't understand why this is here and not inside the node, like LayerPanel or LayerManagement. could you clarify please?

Bluesmile82: Its because of the styling. The margins in the Submenu content will make impossible to show the BasemapPanel as a block with 100% width of the submenu.

PR #757, opened 2017-08-21
Fix/filter group modal dropdowns

Bluesmile82:
image

Fix dropdown style and fix the group layer modal button that was not working


Bluesmile82:
Ready for review

PR #758, opened 2017-08-28
[do not merge] allow not using max zoom from layer headers

nerik:
https://basecamp.com/1756858/projects/12049853/todos/315899461


tiagojsag:
Currently this does not achieve the expected results, as we would need to implement the necessary logic to load and scale data from lower zoom levels when going beyond a layer's configured zoom level.

following up on the discussion here: https://basecamp.com/1756858/projects/12049853/todos/315899461

Issue #645, opened 2017-08-03
determine "regions" and create vector layer


nerik:

nerik:
This could be used for the regions:
![screen shot 2017-08-29 at 14 20 37](https://user-images.githubusercontent.com/1583415/29820642-7d672042-8cc5-11e7-8bf0-919ef2cb364d.png)

(extracted from Skytruth Transshipment report)

nerik:
>  those are the FAO regions, typically the level at which the FAO aggregates their stats. I can share that shapefile tomorrow (I'll check where I got it, possibly Marine Regions). There is also an ocean basins shapefile that I have used in the past to run ocean basin statistics that I'll find. It might be more useful

> Here are the FAO regions shapefile and the world oceans shapefile. The world oceans is from MarineRegions (http://www.marineregions.org/downloads.php#iho). The FAO shapefile is from the UN FAO (http://www.fao.org/geonetwork/srv/en/main.home?uuid=ac02a460-da52-11dc-9d70-0017f293bd28)

PR #761, opened 2017-09-04
Feature/mini globe

Bluesmile82:
This branch depends on feature/fishing-layer as it was needed for the positioning of the panels

image
- [x] Create Mini globe: topoJson and d3-geo
- [x] Rotate Mini gobe on map recenter
- [x] Add Zone marker
- [x] Update zone marker on zoom and window size change

---

The size of the zone marker can be improved across zoom levels but its fair enough. It has a minimum level of display as it won't be really useful if we can see all the world.

The zone is blocked by https://github.com/Vizzuality/GlobalFishingWatch/issues/645


app/src/components/MiniGlobe.jsx
```diff
@@ -0,0 +1,131 @@
...
+class MiniGlobe extends Component {
+
+  constructor(props) {
+    super();
+    this.worldData = feature(jsonData, jsonData.objects.land).features;
+    const viewportWidth = props.viewportWidth;
+    const viewportHeight = props.viewportHeight;
+    const size = this.markerSize(props.zoom, viewportWidth, viewportHeight);
+
+    this.setState({
```
tiagojsag: ?

app/src/components/MiniGlobe.jsx
```diff
@@ -0,0 +1,131 @@
...
+    });
+    this.state = {
+      projection: null,
+      markerWidth: size.width,
+      markerHeight: size.height
+    };
+  }
+
+  markerSize(zoom, viewportWidth, viewportHeight) {
```
tiagojsag: this method is called twice, and each time after it's used you set the result to the state. Perhaps it should be named "setMarkerSize" and it should also update the state.

app/src/components/MiniGlobe.jsx
Bluesmile82: This is the size of the red marker to show the covered map area

PR #763, opened 2017-09-06
Fix/responsive menu fixes

Bluesmile82:
image

- [x] Fix animations
- [x] Scroll Expand Items
- [x] Fix paddings to allow scroll (no negative margins allowed when scroll)
- [x] Fix height of filter items
- [x] Fix info button in add layer modal to make it compatible with the scroll
- [x] Add expand tab for mini globe and layer


tiagojsag:
Please do another pass on this, and make sure it matches the designs. I found a few details that could be improved

app/src/components/MobileLeftExpand.jsx
```diff
@@ -0,0 +1,56 @@
+import React, { Component } from 'react';
```
tiagojsag: Please stop creating new components inside this folder. Move it elsewhere, please

PR #768, opened 2017-09-12
Feature/filter group modal

Bluesmile82:
Add filter group functionality:

- [x] Filter layers country info on the map
- [x] Filter layers gear type info on the map
- [x] Get real headers from the layers
- [x] Save filterGroups to the workspace
- [x] Fix colors
- [x] Filter for every field in the headers (not only gear type id)
- [x] Add deprecated filters from the workspace converting them to group filters with AIS layer
- [x] Update analytics for filters


Bluesmile82:
Please review (Sorry for the big PR)

app/src/actions/heatmap.js
```diff
@@ -324,15 +324,23 @@ export function loadTilesExtraTimeRange() {
...
 }
 
-
+/**
+ * Gets all the categories for all the filters of a Heatmap layer
+ * @param {object} state - the application state
+ * @param {string} layerId - the id of a heatmap layer
+ * @return {array} categories
+ */
 const _getCurrentFlagsForLayer = (state, layerId) => {
```
tiagojsag: we should no longer have flag specific logic on the code

app/src/actions/vesselInfo.js
```diff
@@ -64,11 +64,10 @@ function setCurrentVessel(tilesetId, seriesgroup, fromSearch) {
       if (request.readyState !== 4) {
         return;
       }
-      if (request.status >= 500) {
+      if (request.status >= 500 || request.responseText === 'Not Found') {
```
tiagojsag: if the response text is "Not Found", we should get a matching http code, and test for that.

app/src/analytics/analyticsActions.js
```diff
@@ -77,6 +80,30 @@ export function trackSearchResultClicked(tilesetId, seriesgroup) {
   };
 }
 
+export function trackCreateFilterGroups(filterGroup) {
```
tiagojsag: do not hardcode filters. this should be fully dynamic, based on the header. If the layers change in a couple of weeks, we have to recode the GA tracking, which is a bad way to spend our time

app/src/analytics/analyticsMiddleware.js
```diff
@@ -83,7 +86,25 @@ const GA_ACTION_WHITELIST = [
     type: GA_RECENT_VESSEL_ADDED,
     category: 'Search',
     action: 'Recent Vessels',
-    getPayload: () => 'Selects a recent vessel to view'
+    getPayload: action => action.payload
+  },
+  {
+    type: GA_CREATE_COUNTRY_FILTER,
```
tiagojsag: same as above

app/src/components/Layers/HeatmapSubLayer.js
```diff
@@ -111,6 +112,40 @@ export default class HeatmapSubLayer {
     }
   }
 
+
+  shouldSkipRenderByCategories(frame, index) {
+    // Skip the render by category (country-flag) if the category is not in the tile data
```
tiagojsag: flag specific code, needs to be generic for all filters

app/src/config.js
```diff
@@ -74,6 +74,8 @@ export const TIMELINE_OVERALL_START_DATE_OFFSET = Math.floor(TIMELINE_OVERALL_ST
 // ie how large the clicked region should be for including vessels
 export const VESSEL_CLICK_TOLERANCE_PX = 3.5;
 
+export const AIS_ID = 'fishing2';
```
tiagojsag: you-in-6-months will have no clue what "AIS_ID" means or is used for. Don't be afraid to use long, descriptive variable names. please improve

app/src/constants.js
```diff
@@ -559,3 +559,41 @@ export const FLAGS_SHORTCODES = {
   zm: [127487, 127474],
   zw: [127487, 127484]
 };
+
+export const VMS_REGISTERED_GEARTYPE = {
```
tiagojsag: any way we can load this from the layer headers?

app/src/layers/layersActions.js
```diff
@@ -126,9 +128,18 @@ export function initLayers(workspaceLayers, libraryLayers) {
       .forEach((heatmapLayer) => {
         const headerPromise = loadLayerHeader(heatmapLayer.url, getState().user.token);
         headerPromise.then((header) => {
+          const defaultFilter = {
```
tiagojsag: isn't that the fallback hack we wanna remove? please document it, or we may think it's "real code"

app/src/report/reportActions.js
```diff
@@ -103,9 +103,13 @@ export function sendReport() {
       from: state.filters.timelineInnerExtent[0].toISOString(),
       to: state.filters.timelineInnerExtent[1].toISOString()
     };
-    const currentFlags = state.filters.flags.map(flag => flag.flag)
-      .filter(flag => flag !== undefined)
-      .map(flag => FLAGS[flag]);
+    const currentFlags = state.filterGroups.filterGroups.filter(filter =>
+      filter.filterValues !== undefined &&
+      filter.filterValues.category !== undefined &&
```
tiagojsag: too many comparisons, feels evil. what is this supposed to do?

app/src/layers/layersActions.js
Bluesmile82: Yes, this should be documented

app/src/report/reportActions.js
Bluesmile82: Just map the countries from the filters for the reports. You need to select only the ones that contain a filter values and filtered country. It may be documented or abstracted to other function

app/src/filters/components/FilterGroupForm.jsx
```diff
@@ -62,6 +65,63 @@ class FilterGroupForm extends Component {
...
     filterGroup.filterValues[name] = value;
+
+    this.setState({ filterGroup });
+
+    if (this.state.autoGenLabel) {
+      this.genFilterName();
+    }
+  }
+
+  genFilterName() {
```
Bluesmile82: @tiagojsag Not sure about this: At first is useful but if you change any filter it doesn't update and it could be confusing for the user if he/she forgets to change it, also if we have many filters or the labels are long. We may want to check this with Ariadna.

app/src/report/reportActions.js
Bluesmile82: Ariadna likes the solution but only if we update it when changed

app/src/filters/components/FilterGroupForm.jsx
nerik: Seems logical to me

PR #769, opened 2017-09-18
Feature/speed controls and UI review changes

Bluesmile82:
image
image
image

- [x] Create speed controls to change the speed of the timeline
- [x] Fixes to the timeline highlight and label
- [x] Add badges for vessels and filters in the main menu
- [x] Add search icon to vessel search input
- [x] Vessel info panel is now on every screen
- [x] All the icons in the panels have the same order: pin - color - info
- [x] Info icon triggers the modal directly until we have the polygon select ready
- [x] Fishing hours info button adjusted
- [x] Fix basemap save in workspace
- [x] Remove animation

### DIFF
--- a/app/src/components/Shared/ExpandItem.jsx
+++ b/app/src/components/Shared/ExpandItem.jsx
@@ -4,11 +4,14 @@ import classnames from 'classnames';
 import ExpandItemStyles from 'styles/components/shared/expand-item.scss';
 import { VelocityTransitionGroup } from 'velocity-react';
 
-function ExpandItem({ active, children, accordion }) {
+function ExpandItem({ active, children, accordion, isVesselInfoPanelOpen }) {
   return (
     <div className={ExpandItemStyles._mobileScroll}>
       <div
-        className={classnames(ExpandItemStyles.expandItem)}
+        className={classnames(
+          ExpandItemStyles.expandItem,
+          { [ExpandItemStyles.isVesselInfoPanelOpen]: isVesselInfoPanelOpen }
+        )}
       >
         <VelocityTransitionGroup
           enter={{ animation: 'slideDown', duration: 200, easing: 'easeOutCubic' }}
@@ -31,7 +34,8 @@ function ExpandItem({ active, children, accordion }) {
 ExpandItem.propTypes = {
   active: PropTypes.bool.isRequired,
   accordion: PropTypes.bool,
-  children: PropTypes.node.isRequired
+  children: PropTypes.node.isRequired,
+  isVesselInfoPanelOpen: PropTypes.bool
 };
 
 export default ExpandItem;
--- a/app/src/layers/components/LayerItem.jsx
+++ b/app/src/layers/components/LayerItem.jsx
@@ -75,6 +75,7 @@ class LayerItem extends Component {
   }
 
   render() {
+    // Expandable info panel was removed in 9c74e11
     const { id, hue, reportId, visible } = this.props.layer;
 
     const color = getKeyByValue(COLOR_HUES, hue);
--- a/app/src/layers/components/LayerItem.jsx
+++ b/app/src/layers/components/LayerItem.jsx
@@ -130,17 +130,6 @@ class LayerItem extends Component {
               className={IconStyles.infoIcon}
             />
           </li>
-
-          {/* <li
-            className={LayerItemStyles.itemOptionItem}
-            onClick={() => this.changeExpand('INFO')}
-          >
-            <ExpandItemButton active={this.state.expand === 'INFO'}>
-              <InfoIcon
-                className={IconStyles.infoIcon}
-              />
-            </ExpandItemButton >
-          </li> */}
         </ul>
       );
     }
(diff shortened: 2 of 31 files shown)


app/src/layers/components/LayerItem.jsx
```diff
@@ -124,14 +124,23 @@ class LayerItem extends Component {
...
           <li
             className={LayerItemStyles.itemOptionItem}
+            onClick={() => this.onClickInfo()}
+          >
+            <InfoIcon
+              className={IconStyles.infoIcon}
+            />
+          </li>
+
+          {/* <li
```
nerik: I suggest removing the commented code in a separated commit, with a message explaining why it was removed.
If you'd rather keep it here, please add an explanation along with the commented code.

app/src/layers/components/LayerItem.jsx
```diff
@@ -168,22 +177,22 @@ class LayerItem extends Component {
             />
           </ExpandItem >
         }
-        <ExpandItem active={!layerPanelEditMode && this.state.expand === 'INFO'} arrowPosition={1}>
-          {/* <div className={LayerItemStyles.selectPolygon}>
+        {/* <ExpandItem active={!layerPanelEditMode && this.state.expand === 'INFO'} arrowPosition={1}>
```
nerik: Same.

app/src/mapPanels/rightControlPanel/components/ControlPanel.jsx
```diff
@@ -20,7 +20,6 @@ import LayersIcon from '-!babel-loader!svg-react-loader!assets/icons/layers-menu
 import FiltersIcon from '-!babel-loader!svg-react-loader!assets/icons/filters-menu.svg?name=FiltersIcon';
 import AOIIcon from '-!babel-loader!svg-react-loader!assets/icons/aoi-menu.svg?name=AOIIcon';
 import PinnedVesselList from 'pinnedVessels/containers/PinnedVesselList';
-import Transition from 'react-transition-group/Transition';
```
nerik: Please remove from dependencies.

app/src/mapPanels/rightControlPanel/components/MenuLink.jsx
```diff
@@ -2,13 +2,17 @@ import PropTypes from 'prop-types';
...
+import classnames from 'classnames';
 
-function MenuLink({ title, icon, onClick }) {
+function MenuLink({ title, icon, badge, onClick }) {
   return (
     <div className={MenuLinkStyles.menuLink} onClick={onClick}>
       <div className={MenuLinkStyles.header} >
         <h2 className={MenuLinkStyles.title} >{title}</h2>
-        <div className={ControlPanelStyles.icon} >{icon}</div>
+        <div className={classnames(MenuLinkStyles.rightIcons, { [MenuLinkStyles.hasBadge]: badge > 0 })} >
```
nerik: Let's try to keep the css conventions we have in the project: modifier classes should begin with a _, see https://github.com/Vizzuality/GlobalFishingWatch/pull/703

app/src/timebar/components/SpeedButton.jsx
```diff
@@ -0,0 +1,36 @@
...
+import classnames from 'classnames';
+import iconStyles from 'styles/icons.scss';
+import buttonStyles from 'styles/components/button.scss';
+import SpeedArrowIcon from '-!babel-loader!svg-react-loader!assets/icons/arrow-speed.svg?name=SpeedArrowIcon';
+
+class SpeedButton extends Component {
+
+  changeSpeed() {
+    const ratio = this.props.decrease ? 0.5 : 2;
+    window.speed *= ratio;
```
nerik: Window.speed was a temporary hack, maybe now is a good time to do something better. Do you think we could add the speed param to the app store, so that it could properly be sent to the Timebar container and component?

Additionally:
- speed should be limited (do it in an action or reducer), especially or low speeds where playback could appear 'broken'
- we will need to store the speed in the workspace: the typical use case where speed needed to be customized was typically with sharing a workspace

app/src/components/Shared/ExpandItem.jsx
```diff
@@ -4,11 +4,14 @@ import classnames from 'classnames';
 import ExpandItemStyles from 'styles/components/shared/expand-item.scss';
 import { VelocityTransitionGroup } from 'velocity-react';
```
nerik: Velocity + velocity-react adds a a lot of weight to the bundle size (around 300kb unminified), and it seems we are only using it in one place. Could you please have a look and see if we could easily replace with a simple CSS transition (only if it's super simple to do)

app/src/timebar/components/SpeedButton.jsx
Bluesmile82: Nice comments! I will add it to the store and workspace. I was not sure how will it work from the store at render time

app/src/mapPanels/rightControlPanel/components/MenuLink.jsx
Bluesmile82: In this case is not a modifier, is the container to the icon and badge on the right. Maybe it can use a better name

PR #772, opened 2017-09-21
Fix/ui fixes

Bluesmile82:
- Change stroke color of scroll in timeline
- Fix spaces in date picker input
- Change Vessels activity color and fix margins
- Don't save unpinned Vessels
- info button from vessels selected when we show the vessel info in the panel
- Fix collapse and legend in embed
- Change Vessels menu icon


nerik:
The arrow on the compact version of the location widget should point to the left when expanded.
screen shot 2017-09-21 at 15 39 25

Otherwise, looks good to merge!

PR #774, opened 2017-09-21
How do we store colors in app state?

nerik:
This is a quick hack so that the MPARU color is the same as the one used in the Carto layer:

screen shot 2017-09-21 at 16 41 12

The way we deal with colors in the app state seems pretty fragile to me. How do you feel about removing hue (legacy from old UI) and named colors everywhere and unify using only hex colors strings?
(we'd only need some mechanism to convert hues to colors from a loaded workspace)


Bluesmile82:
I think that would be a nice idea! I don't like having the hue values still. We can select the color with the closest hue loaded from the workspace

PR #786, opened 2017-10-04
Fix/remove legend

Bluesmile82:
Last fixes for the release:

- [x] Remove legend
- [x] Always show option to collapse miniglobe
- [x] Add more colors to the colorpicker
- [x] Adjust searchbox button margins


tiagojsag:
- [ ] Please ask Ariadna to take a look at the design changes here. Feels a bit weird now, but let her decide
- [ ] Colors 1+2 and 3+4 look very very similar between them. Can we make them easier to differentiate? Please test these on the crappiest monitor we have in the office, not all our users will have a Mac with a very nice display

image
image

PR #793, opened 2017-10-09
Fix/toggle color

Bluesmile82:
Fixes the color of the toggles in items which was missing and adds back the AIS layer to the workspace


app/src/layers/components/LayerPanel.jsx
```diff
@@ -72,17 +72,17 @@ class LayerPanel extends Component {
           </ul >
         </ExpandItem >
         <AccordionHeader
-          menuName={'Map Layers'}
```
tiagojsag: Are we sure the client asked us to rename "Map Layers" and not "Fishing Layers"? this doesn't make much sense IMO

tiagojsag:
@Bluesmile82 can't really identify the bug here. Can you explain, please?

app/src/layers/components/LayerPanel.jsx
Bluesmile82: I understood they wanted us to change the name because this layers will not be only fishing layers but also will contain other information like encounters

tiagojsag: Just confirmed with Enrique, to be sure, and it's the other group. In any case, please remove this change from this PR, I will handle it separately

nerik:
@tiagojsag This fixes a bug introduced in develop where toggles would not show the color.

PR #794, opened 2017-10-11
Refacto filters modal

nerik:
- removed duplicated logic in FilterGroupForm. The container now has logic to prepare the data coming from the stores for the component to render.
- made FilterGroupForm a dumb component: it doesn't handle its own state, and directly calls actions on the store for every modification of the currently edited filter.
- creating a new filter group, with initial state, has an explicit action in the actions rather than inside the component
- added support for aliasing filters (ie filters that have the same id but apply to a different field depending on the layers, for example flag --> flag_id or category)
- moved flag options generation code to an util, they are now generated once in a constant
- removed legacy filter code as well as support for headers without filter information (default filter generation)


tiagojsag:
LGTM if my 2 commits dont break anything

PR #810, opened 2017-10-27
Filter group multiselect

nerik:
Adds multiple filter selectors, using react-select, as well as support for this in the heatmap layers.

fishing-multifilter


app/src/filters/filterGroupsReducer.js
```diff
@@ -106,10 +106,10 @@ export default function (state = initialState, action) {
     }
     case SET_CURRENT_FILTER_VALUE: {
       const currentlyEditedFilterGroup = Object.assign({}, state.currentlyEditedFilterGroup);
-      if (action.payload.value === '') {
+      if (!action.payload.values.length || action.payload.values[0].id === '') {
```
tiagojsag: not sure what that [0] means, but are we sure it's not evil?

tiagojsag:
LGTM

PR #819, opened 2017-10-31
Fixed logo (fixes #815)

### LINKED ISSUE #815: Change logo
(From @enriquetuya)
I was just checking the 2.2.1-RC3 on our test environment. This version is the one that contains the new Vizz logo. After trying it out, the one that @ariadna sent back then seems much better than the current one: https://globalfishingwatch.slack.com/files/U6VDT463F/F7P098FCL/02__new_.jpg

Let's move ahead and use that one instead. we already have Paul's approval.

The problems I find with RC3 are:
- Both header and sub-header are not that sharp neither on mobile or desktop.
- Global Fishing Watch is much smaller than the current production logo.
- The vertical allignment of the logo and the text is much better on Ariadna's version

Both issues are fine on the mock that @ariadna sent. 

02 _new

nerik:
screen shot 2017-10-31 at 19 09 07


tiagojsag:
Do we have a svg version of the logo? looks a bit blurry on mobile

nerik:
@tiagojsag Now with SVG logo. Please test on retina.

tiagojsag:
Looks retina ;)
Can you have a look on mobile with sara, please? it seems to not be fully centered, but might just be my eyes. if possible, test on a real phone

PR #832, opened 2017-11-06
Filters description modal

nerik:
Adds a modal displaying the filter description provided by layer headers.
Additionally, puts back landlocked countries to the flags list.


tiagojsag:
any reason why we are putting back the landlocked countries?

nerik:
Yes, that was on Enrique's request (Bolivia is landlocked but has vessels, for instance)

PR #841, opened 2017-11-14
Filter group fixes and layout changes

nerik:
Fixes various issues with filter groups (see corresponding issues) and adds new vertical layout (https://projects.invisionapp.com/share/FUED642GN#/screens/262854072)


app/src/filters/components/FilterGroupItem.jsx
```diff
@@ -48,16 +51,23 @@ class FilterGroupItem extends Component {
       </ul >
     );
 
+    const tooltip = (this.props.filterGroup.label.length > 30) ? this.props.filterGroup.label : null;
```
tiagojsag: I’m wondering if we should make a reusable component out of this “if-short-then-label-else-tooltip” thing.  May be useful for pinned vessel names and such. No need to do it now, just a suggestion. Thoughts?

nerik: Yes probably something like a `<Tooltipable />` higher-order component or the like. Let's keep that in mind.

PR #845, opened 2017-11-23
Update social network card image to new url

tiagojsag:


tiagojsag:
@enriquetuya let us know if you're happy to use this image, or another one

thanks

PR #851, opened 2017-12-04
Feature/scroll bar

Bluesmile82:
- Add new scrollbar useful for Chrome and firefox

image


app/styles/components/shared/react-select.scss
```diff
@@ -90,7 +90,6 @@
     background-color: rgba($color-button-primary, .2);
     border: 0;
     border-radius: 2px;
-    color: $color-white;
```
tiagojsag: this is needed @Bluesmile82

PR #858, opened 2018-02-06
Changes on package.json to reflect repository location change

enriquetuya:
We have moved the GFW map from the Vizzuality organization to GlobalFishingWatch one. We have also migrated the dependent repositories. This PR takes care of the changes needed in the package.json in order for the project to use the correct paths.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,8 @@
     "redux-thunk": "^2.2.0",
     "topojson-client": "^3.0.0",
     "velocity-react": "^1.3.3",
-    "whatwg-fetch": "^2.0.3"
+    "whatwg-fetch": "^2.0.3",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -51,7 +51,8 @@
     "redux-thunk": "^2.2.0",
     "topojson-client": "^3.0.0",
     "velocity-react": "^1.3.3",
-    "whatwg-fetch": "^2.0.3"
+    "whatwg-fetch": "^2.0.3",
+    "globalfishingwatch-convert": "git+https://github.com/GlobalFishingWatch/map-convert.git"
```
tiagojsag: not sure if this belongs here

enriquetuya: Thanks @tiagojsag you are right, that artifact is used in the development branch, not in master. Removed

PR #859, opened 2018-02-07
Fix build

nerik:
- fixes issue with dependency causing an UglifyJS error
- fixes Redux production error message

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "vector-tile": "^1.3.0",
     "velocity-react": "^1.3.3",
     "whatwg-fetch": "^2.0.3",
-    "globalfishingwatch-convert": "git@github.com:GlobalFishingWatch/map-convert.git"
+    "globalfishingwatch-convert": "git+https://github.com/GlobalFishingWatch/map-convert.git"
   },
   "devDependencies": {
     "autoprefixer": "^7.1.2",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "server.js",
   "scripts": {
     "analyze-bundle": "webpack --config ./config/webpack.config.js --json | webpack-bundle-size-analyzer",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -54,7 +54,7 @@
     "vector-tile": "^1.3.0",
     "velocity-react": "^1.3.3",
     "whatwg-fetch": "^2.0.3",
-    "globalfishingwatch-convert": "git+https://github.com/GlobalFishingWatch/map-convert.git"
+    "globalfishingwatch-convert": "git@github.com:GlobalFishingWatch/map-convert.git"
```
enriquetuya: @nerik why did you changed the protocol?

andres-arana: yeah, I don't think we should do this because we run this build process in some GCP machines that are not set up for ssh access.

package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "global-fishing-watch",
-  "version": "2.1.0",
+  "version": "2.4.0",
   "description": "This repository will host the Global Fishing Watch client application developed by Vizzuality",
   "main": "server.js",
   "scripts": {
     "analyze-bundle": "webpack --config ./config/webpack.config.js --json | webpack-bundle-size-analyzer",
-    "build": "webpack --config ./config/webpack.config.js --progress --profile --colors",
+    "build": "NODE_ENV='production' webpack --config ./config/webpack.config.js --progress --profile --colors",
```
andres-arana: When deploying this, we run `NODE_EN=-production npm run build`. Shouldn't this variable be inherited by the webpack process?

nerik: Both work, but in reality AFAIK `build` is never called with anything else than 'production', considering 'development' is just for local testing purposes, and the local environment doesn't use `npm run build` but the webpack dev server.
I'm happy to remove that edit if you prefer.

PR #860, opened 2018-02-23
Workspace get override

nerik:
This implements https://github.com/GlobalFishingWatch/GFW-Tasks/issues/648

It can be tested by appending the following workspace override samples to the client's URL:
```
paramsPlainText={"vessels":[[5572477,"516-resample-v2",45017],[3535204,"516-resample-v2",38829]],"view":[6,-3,48],"innerExtent":[1478707468062,1479469642064],"outerExtent":[1477360207737,1480550400000]}
```
Or, using the base64 encoded version:
```
params=eyJ2ZXNzZWxzIjpbWzU1NzI0NzcsIjUxNi1yZXNhbXBsZS12MiIsNDUwMTddLFszNTM1MjA0LCI1MTYtcmVzYW1wbGUtdjIiLDM4ODI5XV0sInZpZXciOls2LC0zLDQ4XSwiaW5uZXJFeHRlbnQiOlsxNDc4NzA3NDY4MDYyLDE0Nzk0Njk2NDIwNjRdLCJvdXRlckV4dGVudCI6WzE0NzczNjAyMDc3MzcsMTQ4MDU1MDQwMDAwMF19
```
Note: this is encoded using `window.btoa` (not sure what should be used server side).

Not implemented but might be needed ?: adding a `hue` parameter to the vessel tracks.
Also, vessel tracks are set to visible automatically, maybe we want this to be configurable?

### DIFF
--- a/app/src/app/appReducer.js
+++ b/app/src/app/appReducer.js
@@ -1,0 +1,14 @@
+import { SET_IS_EMBEDDED } from './appActions';
+
+const initialState = {
+  isEmbedded: false
+};
+
+export default function (state = initialState, action) {
+  switch (action.type) {
+    case SET_IS_EMBEDDED:
+      return Object.assign({}, state, { isEmbedded: action.payload });
+    default:
+      return state;
+  }
+}
--- a/app/src/components/AuthMap.jsx
+++ b/app/src/components/AuthMap.jsx
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import MapContainer from 'containers/MapContainer';
-import { getURLParameterByName, getURLPieceByName } from 'lib/getURLParameterByName';
+import { getURLParameterByName } from 'lib/getURLParameterByName';
 import Header from 'siteNav/containers/Header';
 import ModalContainer from 'containers/ModalContainer';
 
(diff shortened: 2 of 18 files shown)


app/src/components/AuthMap.jsx
```diff
@@ -24,27 +22,24 @@ class AuthMap extends Component {
...
       <div className="fullHeightContainer" >
-        <Header isEmbedded={this.state.isEmbedded} canShareWorkspaces={canShareWorkspaces} />
-        <ModalContainer
-          isEmbedded={this.state.isEmbedded}
-        />
+        <Header canShareWorkspaces={canShareWorkspaces} />
+        <ModalContainer canShareWorkspaces={canShareWorkspaces} />
         <MapContainer workspaceId={this.state.workspaceId} isEmbedded={this.state.isEmbedded} />
 
       </div >);
```
andres-arana: A bit of weird spacing here?

app/src/index.jsx
```diff
@@ -81,3 +84,6 @@ render(
   </Provider >,
```
andres-arana: Weird spacing here as well

app/src/app/appReducer.js
```diff
@@ -0,0 +1,14 @@
+import { SET_IS_EMBEDDED } from './appActions';
+
+const initialState = {
+  isEmbedded: false
+};
+
+export default function (state = initialState, action) {
+  switch (action.type) {
+    case SET_IS_EMBEDDED:
+      return Object.assign({}, state, { isEmbedded: action.payload });
```
andres-arana: Completely unrelated to this change, but did you know about the object spread operator? You can simplify these kinds of assignments with the equivalent syntax:

```
return {
  ...state,
  isEmbedded: action.payload,
};
```
See https://redux.js.org/recipes/using-object-spread-operator for more examples

nerik: Nice! The object spread operator is used in various places in the codebase, but it never 'clicked' to me that we could use it in Redux reducers.
I'll keep the idea for later as I'd rather have have a refacto to make it consistent (at least within the same reducer)

PR #862, opened 2018-03-13
failsafe to continue loading workspace when timebar data fails

nerik:
I have a CORS issue in production (only on Firefox strangely) with the JSON endpoints for Timebar data. When this fails, the layer library does not get loaded which causes the whole app to crash. This is a structural issue that we need to fix later. 
For now this is a simple hotfix that just loads dummy data.

### DIFF
--- a/app/src/timebar/timebarActions.js
+++ b/app/src/timebar/timebarActions.js
@@ -30,6 +30,17 @@ export function loadTimebarChartData(startDate, endDate) {
         // they should be able to load in parallel but this will need a review
         // of the Timebar component
         dispatch(getLayerLibrary());
+      })
+      .catch(() => {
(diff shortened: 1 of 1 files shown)


enriquetuya:
@nerik what happens if you send and empty array instead of dummy data? I think if the app don't crash that would be preferable, since we don't want anyone think the dummy data is real data.

nerik:
@enriquetuya the D3 line renderer does not like an empty dataset, it has to have at least 2 data points. The dummy data only renders as a solid blue bar, which is not ideal but won't look as real data, just as the absence of data: 

screen shot 2018-03-13 at 19 45 58

nerik:
@andres-arana sure, https://github.com/GlobalFishingWatch/GFW-Tasks/issues/671

PR #866, opened 2018-03-21
Configurable encounters details

nerik:
Makes vessel info fields from the vessel info panel a reusable component. Use it for encounters in order to show all available fields (from the layer's header).

Cancels https://github.com/GlobalFishingWatch/map-client/pull/865
Fixes https://github.com/GlobalFishingWatch/GFW-Tasks/issues/676

There are quite a lot of files changed, as I did a refacto, moving vessel info related files to their own `vesselInfo` namespace.

Logged in:

screen shot 2018-03-21 at 11 57 30

Logged out: 

screen shot 2018-03-21 at 12 10 05


andres-arana:
@nerik the screenhots are reversed, right? logged in it shows more info than logged out, or is there anything I'm missing?

enriquetuya:
@nerik the only comment I have is that we may want to remove the arrow next to the vessel name `->` and replace it with a pin feature.  

The original approach from Vizz I think was to display a summary of the info and then allow the user to see the details if they wanted to. Know, that we display all data, the `->` does not bring any value. Though a user may want to pin the vessel  (this is based on an assumption based on what I experienced loosing the vessel I've just seen in an encounter)

nerik:
@enriquetuya https://github.com/GlobalFishingWatch/GFW-Tasks/issues/682

PR #870, opened 2018-03-27
Do not let KML loading resize viewport except on layer add

### LINKED ISSUE #686: Show active data/ location on search results
https://airtable.com/tbliFUBZhHrWTVy5I/viwjB6S36hfjUz7EU/recbCqzx2wABPdgup

nerik:
Fix for #686.

### DIFF
--- a/app/src/workspace/workspaceActions.js
+++ b/app/src/workspace/workspaceActions.js
@@ -113,7 +113,10 @@ export function saveWorkspace(errorAction) {
       if (newLayer.subtype === LAYER_TYPES.Encounters) {
         newLayer.type = LAYER_TYPES.Encounters;
       }
+      // TODO Should we use a whitelist of fields instead ?
       delete newLayer.header;
+      delete newLayer.justUploaded;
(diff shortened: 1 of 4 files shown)


app/src/workspace/workspaceActions.js
```diff
@@ -113,7 +113,10 @@ export function saveWorkspace(errorAction) {
       if (newLayer.subtype === LAYER_TYPES.Encounters) {
         newLayer.type = LAYER_TYPES.Encounters;
       }
+      // TODO Should we use a whitelist of fields instead ?
```
enriquetuya: @nerik convert this question into a ticket. Don't mind leaving the `TODO`, but if there is not ticket is hard to remember about this unless you step into this line of code again. And when you step into it, you are for sure working in another thing so you can't spend time working on it.

nerik: @enriquetuya Ok. Since it's purely front-end, would you create it in GFW-Tasks?

PR #876, opened 2018-04-17
Send auth header when fetching PBF tiles

nerik:
Fix for https://github.com/GlobalFishingWatch/GFW-Tasks/issues/705


nerik:
As a side note, the reefer layer header gives a 500:
https://api-dot-skytruth-pelagos-production.appspot.com/v2/tilesets/gfw-tasks-658-reefer-uvi-v2/header

PR #883, opened 2018-04-25
Search results fixes

nerik:
Contains the following fixes related to vessel search:
- consistent ordering of search results (fixes https://github.com/GlobalFishingWatch/GFW-Tasks/issues/717)
- fixed an issue where the number of pages available in paginated search results was always higher than needed, resulting in empty pages
- fixed an issue where words were not highlighted when search contained several words (ie a search for "__TUNA QUEEN__" would not result in highlighting "SOME VESSEL WITH THE WORD __QUEEN__")


enriquetuya:
Tested the branch on dev before merging.

PR #903, opened 2018-05-30
Encounters master fixes

nerik:
- shows a "back to transshipment event" button on the vessel info panel
- clicked encounter stays highlighted on the map
- display a small popup over the map on encounters hover


nerik:
https://github.com/GlobalFishingWatch/GFW-Tasks/issues/738
https://github.com/GlobalFishingWatch/GFW-Tasks/issues/740

app/src/components/Map/EncounterInfoWindow.jsx
```diff
@@ -0,0 +1,70 @@
...
+      : (<div
+        className={classnames(
+          CustomInfowindowStyles.customInfowindow,
+          CustomInfowindowStyles._small,
+          CustomInfowindowStyles._topleft
+        )}
+      >
+        <div className={CustomInfowindowStyles.description}>
+          Encounter: {time}<br />
+          {encounter.vessel_1_type} <b>{encounter.vessel_1_information}</b><br />
```
enriquetuya: @nerik make the check if `encounter.vessel_1_information` exist and then display the information. If not only display the time.

public/workspace-encounters.json
```diff
@@ -8,44 +8,33 @@
...
           "opacity": 1,
           "id": "encounters",
-          "tilesetId": "vizz-generated-encounters-pbf-v2"
+          "tilesetId": "vizz-generated-encounters-pbf-v5"
+        },
+        {
+          "title": "AIS Fishing effort",
+          "visible": true,
+          "description": "AIS Fishing effort",
+          "url": "https://api-dot-skytruth-pleuston.appspot.com/v2/tilesets/gfw-tasks-657-uvi-v1",
```
enriquetuya: @nerik we should avoid having workspaces pointing to our dev environment on the repository. We need to plan moving away from this. We can have a separate repo with resources for us to use that has this workspace.  Leave this for now. I'll create a ticket for later.

nerik: @enriquetuya I had git <a href="https://github.com/GlobalFishingWatch/map-client/blob/develop/.gitignore#L24">ignoring</a> that workspace - as its only use is for local dev. I also removed the reference to a dev env tileset that was in the default local workspace (workspace.json)

nerik:
@enriquetuya 
Re https://github.com/GlobalFishingWatch/GFW-Tasks/issues/742

- fixed the search issue that was caused by the fact that the `searchable` flag was not implemented by the client side yet

- fixed lack of scrollbar in layer info modal

- `There are multiple vessels at this location` ->  `objects`

- z-order: this was actually working, as in last layer declared in workspace goes on top. The issue is that all activity layers use a specific 'screen' pixel blend mode (a blend mode defines how pixels are combined), which gives that 'blooming' effect to the heatmap.
I added a rule so that the encounter layer uses a normal blend mode, which results in this 👍 

screen shot 2018-06-01 at 12 42 59

Before:
screen shot 2018-06-01 at 12 43 23

The VIIRS layer visual aspect has not changed. My suggestion (and I should have thought of that before) is to have a **distinct visual style for VIIRS points**, as heatmap style does not work for this dataset: maybe a more adequate style could use plain simple circles, similar to the heatmap zoomed-in style.

- layers that have `pinable: false` in their header will now be prevented from being pinned.

- `vessel_n_information` is now only displayed in encounters hover popup when available (so just the date for now). We could have the duration too but that requires a modification on the layer header. 

- search now only happens on added and **visible** layers. If there are searchable but invisible layers the following message is displayed:

screen shot 2018-06-01 at 11 05 33

enriquetuya:
I have tested all changes in version 2.5.0-rc4. Also I've gone through the PR besides the comment on the workspace having mention to dev environment tiles everything seems good.

PR #904, opened 2018-06-11
Encounters: use workspace layer id as a fallback to layer tilesetId

nerik:
Fix for https://github.com/GlobalFishingWatch/GFW-Tasks/issues/749

### DIFF
--- a/app/src/encounters/encountersActions.js
+++ b/app/src/encounters/encountersActions.js
@@ -77,9 +77,11 @@ export function setEncountersInfo(seriesgroup, tilesetId) {
       const workspaceLayers = getState().layers.workspaceLayers;
 
       encounterInfo.vessels.forEach((vessel) => {
-        const workspaceLayer = workspaceLayers.find(layer => layer.tilesetId === vessel.tilesetId);
-        const fields = workspaceLayer.header.info.fields;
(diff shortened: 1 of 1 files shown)


app/src/encounters/encountersActions.js
```diff
@@ -77,8 +77,9 @@ export function setEncountersInfo(seriesgroup, tilesetId) {
       const workspaceLayers = getState().layers.workspaceLayers;
 
       encounterInfo.vessels.forEach((vessel) => {
-        const workspaceLayer = workspaceLayers.find(layer => layer.tilesetId === vessel.tilesetId);
+        const workspaceLayer = workspaceLayers.find(l => l.tilesetId === vessel.tilesetId || l.id === vessel.tilesetId);
```
enriquetuya: @nerik I  don't like variable names that are not self_descriptive and 1 character of length. Could you please change the name of the variable? Besides that the change looks good.

andres-arana: @enriquetuya @nerik while I would generally agree with you @enriquetuya, in lambda functions it's usually ok. Usually, when you have a single-argument lambda, it's common to use `it` as the argument name, as the lambda is already descriptive enough.

nerik: @andres-arana I think I fixed while you were writing your comment :p

PR #942, opened 2018-08-14
High zoom levels handling

nerik:
Fix for https://github.com/GlobalFishingWatch/map-client/issues/940

Previous behaviour:
User could not zoom beyond z level 10. This value was set on each activity layer header (`max_zoom`) and apparently was always the same for all layers. The client did not correctly handle different zoom levels anyways.

Now:
User can zoom up to a more detailed z level, set arbitrarily as 14.
Removed faulty z level logic, maximum zoom level for loading of tiles is hardcoded at 10. For z levels 11-14, the map will still display activity layers, but with the last available z level (10)

@enriquetuya let me know if per-layer max zoom level is actually required (considering we never had this working on the client side and that it's not trivial to implement)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 3.0.0
+- Deeper zoom levels up to 14, even when activity layer tiles are not visible (https://github.com/GlobalFishingWatch/map-client/pull/942)
+
 ## 2.4.0 RC2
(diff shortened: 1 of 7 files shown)


nerik:
@enriquetuya No impact on tracks as they are independant from z level.

Unrelated, there are several issues with workspace saving/loading, do not merge yet.

PR #943, opened 2018-08-15
Workspace gl layers

nerik:
This adds the possibility to directly specify Mapbox GL JSON for workspace layers, allowing pulling layers from CARTO without having to have them specified in the `directory` endpoint nor in the base GL JSON style (use cases described here: https://github.com/GlobalFishingWatch/map-client/issues/932#issuecomment-412865322)

The gist of it is: 
- set up a CARTO table, have it public
- add the SQL query and a bunch of other parameters in the workspace layer `gl` parameter.
This is fully documented, with examples, here: https://github.com/GlobalFishingWatch/map-client/blob/a0e6283d539c75048a4cfa9221797f50d14b9b2b/documentation/workspaces.md
This is the workspace used for tests: https://github.com/GlobalFishingWatch/map-client/blob/a0e6283d539c75048a4cfa9221797f50d14b9b2b/public/workspace.json

@enriquetuya I made that choice because:
- this avoids having any conversion/parsing step on the client side code, reducing complexity (GL JSON is just taken from the workspace layer and merged with the base GL JSON style)
- it doesn't require creating a CARTO map/vizjson, only a public table and a SQL query are required.
- this gives complete control over the appeareance and behavior of the custom layer: SQL, popups, labels, various styling attributes, etc.
- from there, allowing custom raster and/or WMS workspace layers should be a breeze.

The disadvantage is that it is slightly more complex than it could be, so let me know if I should clarify anything.

### DIFF
--- a/documentation/workspaces.md
+++ b/documentation/workspaces.md
@@ -2,6 +2,118 @@
 
 ## Workspace parameters
 
+### map:center
+
+A lat, lon pair for the map center.
+
+### map:zoom
+
+The zoom value as a float (non integer values are allowed).
+
+### map:layers
+
+#### type
+
+- `CartoDBAnimation`: static, vector based layers served from a CARTO instance, and rendered by Mapbox GL.
+- `ClusterAnimation`: activity layers, rendered by Pixi.
+
+#### opacity
+
+#### color
+
+#### visible
+
+#### gl
+
+The `gl` parameter allows specifying custom Mapbox GL JSON layers, even if they are not set in the <a href="https://github.com/GlobalFishingWatch/map-client/blob/develop/app/src/map/gl-styles/style.json">base GL JSON</a> style. 
+
+`gl` needs two mandatory parameters:
+- `source`: A source can set a SQL query on a CARTO table <a href="https://github.com/GlobalFishingWatch/map-client/blob/develop/documentation/layers.md#gfwcarto-sql">as described in the layers documentation</a>, as well as the fields displayed in popups. Source id will automatically be set to the workspace layer id.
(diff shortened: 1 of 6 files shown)


public/workspace.json
```diff
@@ -2,18 +2,11 @@
   "workspace": {
```
enriquetuya: @nerik are the changes in this file intended? maybe we want to have a `galery/` folder and include `custom-gl-layer-workspace.json`?

documentation/workspaces.md
```diff
@@ -2,6 +2,118 @@
...
+  "title":	"South American countries",
+  "type": "CartoDBAnimation",
+  "id": "samerica_adm0",
+  "color": "#ff00ff",
+  "visible": true,
+  "gl": {
+    "source": {
+      "type": "vector",
+      "metadata": {
+        "gfw:carto-sql": "SELECT * FROM samerica_adm0",
```
enriquetuya: @nerik we only support queries to our carto instance? where is that configured?

nerik: @enriquetuya This is hardcoded here for now: https://github.com/GlobalFishingWatch/map-client/blob/develop/app/src/config.js#L205
Should I make this an env variable?

public/workspace.json
nerik: @enriquetuya This is useful just for testing this PR. Maybe this shouldn't even be versioned? I like your idea, but that means maintaining those workspaces...

documentation/workspaces.md
andres-arana: +1 to making this configurable in an env file.

PR #944, opened 2018-08-15
multiple tilesets search endpoint

nerik:
Currently blocked by two issues on the API.
- I can't get multi tileset search to work.
Single tileset returns 37 results:
`https://api-dot-skytruth-pleuston.appspot.com/v2/tilesets/gfw-tasks-657-uvi-v2/search/?query=bol&limit=14&offset=0`
Same query, 2 tilesets give 0 results:
`https://api-dot-skytruth-pleuston.appspot.com/v2/tilesets/gfw-tasks-657-uvi-v2,indo-public-fishing/search/?query=bol&limit=14&offset=0`
- we're going to need to integrate the `tilesetId` in each search result, in order to call the correct endpoint when clicking on each search result.

### DIFF
--- a/app/src/search/searchActions.js
+++ b/app/src/search/searchActions.js
@@ -29,11 +29,7 @@ const loadSearchResults = debounce((searchTerm, page, state, dispatch) => {
 
   const searchableLayers = state.layers.workspaceLayers
     .filter(layer => LAYER_TYPES_SEARCHABLE.indexOf(layer.subtype || layer.type) > -1)
-    .filter(layer =>
-      layer.header.searchable !== false &&
-      layer.header.endpoints.search !== null &&
-      layer.header.endpoints.search !== undefined
-    )
+    .filter(layer => layer.header.searchable !== false)
     .filter(layer => layer.added === true);
 
   const searchableAndVisibleLayers = searchableLayers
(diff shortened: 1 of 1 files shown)


nerik:
https://github.com/GlobalFishingWatch/GFW-Tasks/issues/718

enriquetuya:
@nerik @andres-arana the problem was that `indo-public-fishing` didn't have the `searchable: false` flag in the header. That layer is not searchable. Please try again now.

enriquetuya:
@nerik @andres-arana in terms of:
Erik:
>we're going to need to integrate the tilesetId in each search result, in order to call the correct endpoint when clicking on each search result.

I think we also need to reference the tileset name in the results in some way, what do you think? That way you know where the result is from.

nerik:
@enriquetuya @andres-arana 
- the multisearch issue is not happening again. Note that the `searchable` flag was, and still is, taken into account (but it must be explicitely set to `false`, if the value is omitted it is considered to be `true`, which maybe is counterintuitive).

- I have added client side code that assumes that `tilesetId` is present on search `entries` items. It fall backs on hardcoded `'gfw-tasks-657-uvi-v2'` so I could test it, we'll need to remove that.
https://github.com/GlobalFishingWatch/map-client/blob/7b1291815b868eaec152fda57f0a6e16108361e6/app/src/search/searchActions.js#L86

- 
> I think we also need to reference the tileset name in the results in some way, what do you think? That way you know where the result is from.

Good idea. I have created an issue here: https://github.com/GlobalFishingWatch/map-client/issues/949

PR #966, opened 2018-09-18
Automatically open right panel when clicking on a vessel

nerik:
Fix for issue explained here:
https://github.com/GlobalFishingWatch/map-client/issues/941

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,16 @@
 # Changelog
 
+## 3.0.0
+- Allow custom Mapbox GL JSON style in workspaces - [#943](https://github.com/GlobalFishingWatch/map-client/pull/943)
+- Deeper zoom levels up to 14, even when activity layer tiles are not visible - [#942](https://github.com/GlobalFishingWatch/map-client/pull/942)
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -3,7 +3,9 @@
 - Added a button to fit the viewport and timebar to vessel tracks
```
enriquetuya: @nerik all the changes in the changelog should go under the `3.0.0` title.

CHANGELOG.md
```diff
@@ -3,7 +3,9 @@
...
 - Removed automatic fit to viewport and timebar when clicking on a vessel search result (can now be done by explicitely clicking the button)
 - Clicking the (i) vessel button again now hides vessel details
-- Harmonisation of icons
+- Harmonisation of icons and various samll UI fixes
+- Right panel automatically expands when clicking on a vessel on the map
```
enriquetuya: @nerik another suggestion for the changelog is that we should start using the same format we have in other projects:
`pullrequest-link - Description`, that way a user interested in knowing how a given change was done can easily navigate to the PR. You can check the PIPE-[x] pr repos to see that structure.

PR #1003, opened 2019-01-07
Fixed error in layers names in WMS GetMap query

nerik:
`GetMap` actually requires layer names, not layer indexes, which for some reason still worked with our test WMS endpoint.


enriquetuya:
Relates to https://github.com/GlobalFishingWatch/map-client/pull/1003

PR #1008, opened 2019-01-31
Wms multi layer

j8seangel:
This PR includes support to select capabilities in WMS layers to fix https://github.com/GlobalFishingWatch/GFW-Tasks/issues/946

kapture 2019-01-31 at 18 46 36

The workflow has changed slightly so now then proxy request creation is done in url changes (also validating a valid url) and the layer is added on done button click.

To finish:
- [x] Remove sub-layers on layer removal.

Bonus:
- Vscode allows you to [resolve imports](https://github.com/GlobalFishingWatch/map-client/compare/wms-multi-layer?expand=1#diff-9be4f8e03a4f3bffa2e1404410a10da7R1) clicking on the name when pushing `cmd`.
- [Redux dev tool](https://github.com/GlobalFishingWatch/map-client/compare/wms-multi-layer?expand=1#diff-903b1079d79b48c21015724a6fff2aa4R84) in developmment, however needed to skip some [store serializations](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/Troubleshooting.md#excessive-use-of-memory-and-cpu) to avoid out of memory crashes.


app/src/layers/containers/CustomLayer.js
```diff
@@ -1,22 +1,40 @@
 import { connect } from 'react-redux';
 import CustomLayer from 'layers/components/CustomLayer';
 import { setLayerManagementModalVisibility } from 'app/appActions';
-import { uploadCustomLayer } from 'layers/customLayerActions';
+import { resetCustomLayerForm, uploadCustomLayer, confirmCustomLayer } from 'layers/customLayerActions';
 import { login } from 'user/userActions';
 
+const getSubLayers = ({ capabilities }) => {
+  const layers = capabilities && capabilities.Capability && capabilities.Capability.Layer && capabilities.Capability.Layer.Layer;
```
nerik: IMHO the container should not have to deal with WMS-specific stuff. This belongs to the custom layer reducer or actions.

jsconfig.json
```diff
@@ -0,0 +1,44 @@
+{
```
nerik: Nice 👌
See https://github.com/GlobalFishingWatch/map-client/issues/1009

enriquetuya:
@nerik @j8seangel 

`User should not be able to validate when 0 WMS sublayers are selected.`
- I didn't get what you are saying here. If you mean that if no sublayers are present, display the whole thing. I buy that. 

`All sublayers are selected by default. My assumption is that in most cases, a user only wants one or a few sublayers, which makes WMS endpoint with 40 sublayers very annoying to use. Either uncheck all by default or offer a "select all / deselect all" feature`
- I would uncheck all by default.

`Some descriptions can be very long and you were right to shorten them, but my assumption is that in some cases users need to be able to read them entirely. Tooltip? Modal opening a modal ?`
- I would add `...` and when you hover that display everything in a tooltip. 

`IMHO allow users to edit sublayer selection after creation is overkill, but we should include in the layer description a list of included sublayers for reference (visible later by clicking (i) in the layers panel)`
- I agree with Erik, to edit delete the layer you added and then add it again. If we get to many complaints we can revisit. 

`sublayers preview, as you suggested initially? Could be done by fetching the 0,0,0 tile I guess.` 
- Unless it is really easy to do, I would avoid the preview. It is difficult to get that right.

app/src/layers/containers/CustomLayer.js
j8seangel: You were right, this is now done in actions.

nerik:
LGTMTM 🚀

PR #1013, opened 2019-02-14
Vessel warning notification

j8seangel:
- Move notification to a redux module to be able to dispatch notifications from any place of the app.
- Add types to existing notifications and sort by priority
- Show warning notification when a vessel Info contains `comment` in the response
- Add warning icon in pinned and detailed vessel area.

image

**TODO**: 

- [x] [Remove mocking code](https://github.com/GlobalFishingWatch/map-client/compare/vessel-warning-notification?expand=1#diff-79c70c2d0b094b469b73c774b9c211beR109) when API ready

**To Discuss**:
I've created a jsx component in the container element to have:
- Presentational component isolated.
- Custom logic onCloseClick out of the presentational component.
- ComponentDidMount lifecycle to dispatch the existing first notification. what are your thoughts [here](https://github.com/GlobalFishingWatch/map-client/compare/vessel-warning-notification?expand=1#diff-11101aec2a341228942e87fffdfd60f1R97) hopefully could be replaced with a simpler [hook API](https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect) after upgrade React version.


app/src/vesselInfo/vesselInfoActions.js
```diff
@@ -96,39 +98,20 @@ function setCurrentVessel(tilesetId, seriesgroup, fromSearch) {
...
+    fetchEndpoint(vesselInfoUrl, token).then((data) => {
       delete data.series;
-
       data.tilesetId = tilesetId;
 
+      // mocking API comments to show warning
+      // TODO:
+      //  1. Check field name from API
+      //  2. Remove the next line of code
+      data.comment = 'This vessel has a warning';
```
enriquetuya: @j8seangel I would remove this now so we can merge the PR, once the API starts delivering the comment it should work.

app/src/vesselInfo/vesselInfoActions.js
```diff
@@ -265,48 +254,23 @@ export const applyFleetOverrides = () => (dispatch, getState) => {
...
-        }
-        const data = JSON.parse(request.responseText);
+      const pinnedVesselUrl = buildEndpoint(layer.header.endpoints.info, { id: pinnedVessel.seriesgroup });
+      fetchEndpoint(pinnedVesselUrl, token).then((data) => {
         delete data.series;
+        // mocking API comments to show warning
+        // TODO:
+        //  1. Check field name from API
+        //  2. Remove the next line of code
+        data.comment = 'This vessel has a warning PINNED';
```
enriquetuya: @j8seangel I would remove this now so we can merge the PR, once the API starts delivering the comment it should work.

public/literals.json
```diff
@@ -7,5 +7,6 @@
   "webgl_warning": "⚠️ There is a problem with your current configuration (WebGL is disabled or unavailable). The map won't be able to load.",
   "deprecated_layers_warning": "The activity layer $LAYERS used on this workspace is deprecated and may not get updated data, click <a href='http://globalfishingwatch.org/map'>here</a> to get the latest one.",
   "deprecated_layers_warning_plural": "The activity layers $LAYERS used on this workspace are deprecated and they may not get updated data, click <a href='http://globalfishingwatch.org/map'>here</a> to get the latest ones.",
-  "edge_warning": "Map performance might be degraded using Edge. We recommand using another browser such as Firefox or Chrome."
+  "edge_warning": "Map performance might be degraded using Edge. We recommand using another browser such as Firefox or Chrome.",
+  "vessel_warning": "This vessel has comments, please review them"
```
enriquetuya: @j8seangel where is this key being used? I've search the PR looking for `vessel_warning` and the one in literals.json is the only one that appeared.

enriquetuya: @j8seangel I would rephrase to `Our analysts have reviewed this vessel and added some comments about it's behavior, please review them in the vessel info.`

app/src/vessels/components/Vessel.jsx
```diff
@@ -62,6 +63,11 @@ class Vessel extends Component {
         >
           {vessel.title}
         </div>
+        {hasWarning &&
+          <div onClick={() => this.props.showWarning(vessel.comment)}>
+            <IconButton icon="alert" />
```
enriquetuya: @j8seangel @satellitestudiodesign shouldn't the warning icon be displayed in the same color as the banner on top?

app/src/notifications/components/notifications.scss
```diff
@@ -43,3 +38,47 @@
     }
   }
 }
+
+.notification {
+  background-color: $color-19;
```
enriquetuya: @j8seangel @satellitestudiodesign shouldn't the background of the notification have less opacity? It feels weird to have such a solid color, at least based on the  image included in the PR description.

satellitestudiodesign: @enriquetuya We included the differentiation between notifications, warnings, and errors precisely to grab users attention when reading the message is mandatory. That's why I went with solid, strong color blocks.

app/src/vessels/components/Vessel.jsx
satellitestudiodesign: @enriquetuya adding contextual colors to icons is something that I like and I'm including in the new version of the UI. Currently, the UI only uses color to tell users what things are interactive, so I believe we should keep it consistent

public/literals.json
satellitestudiodesign: @j8seangel @enriquetuya My idea was to show the analyst's comment directly on the warning, so I thought this message was temporary. We can add it to the vessel info as well, I like that.

j8seangel: Sorry, this is my fault and I forgot to add it as the title of the warning icon in the vessel info.

Just to clarify the workflow when a vessel info request has the comment field on the response:
- The warning notification bar will be shown the content of the comment itself automatically.
- Vessel info and the pinned vessel sections will show a warning icon with the above comment as a title that triggers the notification with the comment as a warning.

app/src/vesselInfo/vesselInfoActions.js
j8seangel: Sounds, good, it is there to test the feature so I will remove it before merging.

app/src/components/App.jsx
```diff
@@ -1,22 +1,16 @@
 /* eslint-disable react/no-danger */
 import React, { Component } from 'react';
-import * as PIXI from 'pixi.js';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import AppStyles from 'styles/components/app.scss';
-import BannerStyles from 'styles/components/banner.scss';
 import ReactGA from 'react-ga';
-import platform from 'platform';
+import Notifications from 'src/notifications/containers/Notifications';
```
nerik: 👍 
later: https://github.com/GlobalFishingWatch/map-client/issues/1009

app/src/notifications/containers/Notifications.js
```diff
@@ -0,0 +1,112 @@
...
+    const element = config[i];
+    if (element.checker()) {
+      return element;
+    }
+  }
+  return null;
+};
+
+const mapStateToProps = (state) => {
+  const visible = state.notifications.visible && window.innerWidth > 768;
```
nerik: I think we need to address that, if it's not too complictaed. This was a quick hack at the time I implemented this notification but the error messages should also be displayed on mobile. cc @satellitestudiodesign

j8seangel:
Fixes https://github.com/GlobalFishingWatch/GFW-Tasks/issues/972

j8seangel:
All comments addressed @enriquetuya, thanks for remember me to update the changelog!

app/src/notifications/containers/Notifications.js
```diff
@@ -0,0 +1,112 @@
...
+    content: state.notifications.content,
+    initialNotification: notificationsConfig && checkInitialNotification(notificationsConfig)
+  };
+};
+
+const mapDispatchToProps = dispatch => ({
+  setNotification: payload => dispatch(setNotification(payload))
+});
+
+class NotificationsContainer extends React.Component {
```
nerik: Why not having the component as a separate `Notifications.jsx` file?

app/src/notifications/notificationsActions.js
```diff
@@ -0,0 +1,19 @@
+export const SET_NOTIFICATION = 'SET_NOTIFICATION';
+
+export function resetNotification() {
```
nerik: Any reason not to use a more modern form?
```
export const resetNotification = () => ({
    type: SET_NOTIFICATION,
    payload: {
      visible: false,
      content: '',
      type: 'notification'
    }
  });
```

app/src/vesselInfo/vesselInfoActions.js
```diff
@@ -96,39 +98,20 @@ function setCurrentVessel(tilesetId, seriesgroup, fromSearch) {
...
-      if (request.readyState !== 4) {
-        return;
-      }
-      if (request.status >= 500 || request.status === 404) {
-        console.error('Error loading vessel info:', request.responseText);
-        return;
-      }
-      const data = JSON.parse(request.responseText);
+    const vesselInfoUrl = buildEndpoint(layer.header.endpoints.info, { id: seriesgroup });
+    fetchEndpoint(vesselInfoUrl, token).then((data) => {
```
nerik: Thanks 👍

app/src/vesselInfo/vesselInfoActions.js
```diff
@@ -395,8 +358,11 @@ export function addVesselFromEncounter(tilesetId, seriesgroup) {
 }
 
 export function clearVesselInfo() {
-  return {
-    type: CLEAR_VESSEL_INFO
+  return (dispatch) => {
+    dispatch({
+      type: CLEAR_VESSEL_INFO
+    });
+    dispatch(resetNotification());
```
nerik: Won't this close _any_ notification? That could close a notification that has nothing to do with the vessel, right?

app/styles/_settings.scss
```diff
@@ -89,6 +89,7 @@ $color-35: #153b78;
 $color-36: #0A1E50;
 $color-37: #ffbcc6;
 $color-38: rgba(22,48,103,0.2);
+$color-39: #ffcc6e;
```
nerik: I'm really sorry 😢

app/src/vesselInfo/vesselInfoActions.js
j8seangel: You are right, but for now we only have that notification apart from the initial one (that will be replaced for the warning when occurs)
So we could include a system to support multiple notifications but I think doesn't worth it for now.

app/src/vesselInfo/vesselInfoActions.js
j8seangel: My pleasure

app/src/notifications/containers/Notifications.js
j8seangel: Do you mean to have `NotificationsContainer.jsx` and and `Notifications.jsx` ? Or have both in the same file ?

app/src/notifications/notificationsActions.js
j8seangel: Nothing special... sorry as I always forget it! 
Already updated

app/src/notifications/containers/Notifications.js
j8seangel: Fixed with some styles review https://github.com/GlobalFishingWatch/map-client/pull/1013/commits/54bca9b58bd4cd1d542fa9b44053b3dc0820a98b

PR #1019, opened 2019-03-01
Include bablic script with language selector

j8seangel:
Branch to test [bablic](https://www.bablic.com/adminpanel/) integration as It runs in local with a warning. But to be able to update config in bablic dashboard we have to publish to staging or production first.

image

### DIFF
--- a/app/src/map/components/Map.jsx
+++ b/app/src/map/components/Map.jsx
@@ -22,11 +22,6 @@ class Map extends React.Component {
     // there is a problem with the container width computation (only with "fat scrollbar" browser/os configs),
     // seems like the panels with scrollbars are taken into account or smth
     window.setTimeout(() => this._resize(), 10000);
-
-    // Shows the language selector hidden until map load finish
-    if (bablic !== undefined && bablic.widget && typeof bablic.widget.show === 'function') {
-      bablic.widget.show();
-    }
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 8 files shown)


app/src/map/components/Map.jsx
```diff
@@ -22,6 +22,11 @@ class Map extends React.Component {
     // there is a problem with the container width computation (only with "fat scrollbar" browser/os configs),
     // seems like the panels with scrollbars are taken into account or smth
     window.setTimeout(() => this._resize(), 10000);
+
+    // Shows the language selector hidden until map load finish
+    if (bablic !== undefined && bablic.widget && typeof bablic.widget.show === 'function') {
```
nerik: I think this should be somewhere else, as `<Map />` is part of the map module, not the app where translations should happen

j8seangel: Indeed, this needs to be moved when `as-module` branch goes to production, but for now, I thought it could fit here.

nerik:
I think we should implement the language selector in a "stable" place where it doesn't have to appear and disappear:

screenshot 2019-03-01 at 15 00 41

This would need a custom selector but I think it's worth it.

Also, where is french?

j8seangel:
Added new <LanguageSelector /> component to customize styles.
kapture 2019-03-04 at 13 46 22

PR #1020, opened 2019-03-11
Bablic fix

j8seangel:
To avoid crashes this PR check when bablic was loaded to request and render languages selector.

Bonus: order languages by key alphabetically to ensure the same order on change.

### DIFF
--- a/app/src/components/Shared/LanguageSelector.jsx
+++ b/app/src/components/Shared/LanguageSelector.jsx
@@ -17,7 +17,7 @@ class LanguageSelector extends Component {
   };
 
   componentDidMount = () => {
-    if (window.bablic !== undefined && bablic.loaded) {
+    if (window.bablic !== undefined && window.bablic.loaded) {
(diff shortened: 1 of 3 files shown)


app/src/components/Shared/LanguageSelector.jsx
```diff
@@ -16,46 +17,56 @@ class LanguageSelector extends Component {
   };
 
   componentDidMount = () => {
-    if (window.bablic !== undefined) {
-      const languages = window.bablic.languages.get();
-      const currentLng = window.bablic.getLocale();
+    if (window.bablic !== undefined && bablic.loaded) {
```
nerik: `window.bablic.loaded`
In fact I think we should have an eslint rule for this.

j8seangel:
```suggestion
    if (window.bablic !== undefined && window.bablic.loaded) {
```

PR #1021, opened 2019-03-12
Map as component integration

j8seangel:
Use the map module from the map-components repo
> Working only locally with `yarn link` 

TODO before release:
- [x] Publish new `map-components` version when module tested and ready.
- [x] Update package.json with new version


j8seangel:
I wasn't able to include pixi.js and react-map-gl within the bundler... so they are still app dependencies. We can improve it later, but at least it is working for now.

Rest were moved and package.json cleaned, remember to run yarn unlink "@globalfishingwatch/map-components" and yarn install

PR #1026, opened 2019-04-11
Map dynamic import

j8seangel:
See https://github.com/GlobalFishingWatch/map-components/pull/43, to improve also get the webGL function checker from pixi repo and copy into utils, this avoids import the whole library just to check it.

Release map-components and update version before merging 🙏 

Before (white screen for longer):
Kapture 2019-04-11 at 12 52 12

Now:
Kapture 2019-04-11 at 12 49 19

### DIFF
--- a/src/app/map/components/MapWrapper.jsx
+++ b/src/app/map/components/MapWrapper.jsx
@@ -2,12 +2,23 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import StaticLayerPopup from 'app/map/components/StaticLayerPopup'
 import HoverPopup from 'app/map/components/HoverPopup'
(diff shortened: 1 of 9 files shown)


src/app/map/components/MapWrapper.jsx
```diff
@@ -2,12 +2,23 @@ import React, { Component } from 'react'
...
     clickPopupData: null,
+    MapModule: null,
+  }
+
+  componentDidMount() {
+    this.loadMapModule()
+  }
+
+  loadMapModule() {
+    import('@globalfishingwatch/map-components/components/map').then((map) => {
```
nerik: 🤯

nerik:
Awesome 🙇‍♂️

PR #1029, opened 2019-05-17
Encounters mapboxgl

nerik:
Adds support for a Mapbx GL powered encounters layer.
Needs https://github.com/GlobalFishingWatch/map-components/pull/56

Although this is experimental, I'd like to merge as it fixes several issues.

The workspace layer that needs to be used:
```
        {
          'id': 'encounters_ais',
          'visible': true,
          'interactive': true,
          'color': '#eeff00',
          "url": "https://api-dot-skytruth-pelagos-production.appspot.com/v2/tilesets/gfw-task-673-encounters-v3",
          "type": "CartoDBAnimation",
          "label": "Encounters",
        },
```

We will need to generalize the behaviour created for encounters to all interactive time-series layer based on Mapbox GL:
- loading of headers 
  - needs adding an headerURL field to workspace.json
  - refactor workspace format entirely now? (https://github.com/GlobalFishingWatch/map-client/issues/871)
- refactor interaction.actions
  - interaction priority should not always be given to heatmap 
  - object returned to callback needs to be more useful
- stop using the `type` field on workspace layer to determine in which Panel the layer will go

### DIFF
--- a/src/app/layers/layersActions.js
+++ b/src/app/layers/layersActions.js
@@ -36,8 +36,10 @@ function loadLayerHeader(tilesetUrl, token) {
     headers.Authorization = `Bearer ${token}`
   }
 
+  const headerUrl = `${tilesetUrl}/header`
+
   return new Promise((resolve) => {
-    fetch(`${tilesetUrl}/header`, {
+    fetch(headerUrl, {
       method: 'GET',
       headers,
     })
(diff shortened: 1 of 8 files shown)


src/app/layers/layersActions.js
```diff
@@ -177,7 +179,11 @@ export function initLayers(workspaceLayers, libraryLayers) {
     // get header promises
     const headersPromises = []
     workspaceLayers
-      .filter((l) => LAYER_TYPES_WITH_HEADER.includes(l.type) && l.added === true)
+      .filter(
+        (l) =>
+          (l.id === 'encounters_ais' || LAYER_TYPES_WITH_HEADER.includes(l.type)) &&
```
j8seangel: Why not move this to a constant? as it is used also here https://github.com/GlobalFishingWatch/map-client/pull/1029/files#diff-409f4d26ff67e6335b043fc86967e616R10 and several places it could be an issue in the future

src/app/map/containers/MapWrapper.js
```diff
@@ -95,23 +95,28 @@ const getHeatmapLayers = createSelector(
         }
         return layerParams
       })
+    console.log(heatmapLayers)
```
j8seangel: 👀 🔴

j8seangel:
Looks good, but please ensure not merge before releasing and merging https://github.com/GlobalFishingWatch/map-components/pull/56

PR #1032, opened 2019-05-27
refacto layer interaction organization / move temporal layers to Mapbox GL / Chile layers

nerik:
Leverages https://github.com/GlobalFishingWatch/map-components/pull/57 for finer control of how user interacts with maps when several layer overlaps.

Also finalizes move of non heatmap temporal layers (encounters + Chile layers) to Mapbox GL

TODO:
- [x] definitely remove support for Pixi encounters layers
    - [x] convert existing workspace layers
- [x] Fixed issue with overlapping Pixi heatmaps
- [x] Chile layers
  - [x] debug
  - [x] visual appearance


src/app/search/searchActions.js
```diff
@@ -61,43 +59,88 @@ const loadSearchResults = debounce((searchTerm, page, state, dispatch) => {
...
-  const baseURL = `${
-    process.env.REACT_APP_V2_API_ENDPOINT
-  }/tilesets/{{tilesets}}/search/?query={{query}}&limit={{limit}}&offset={{offset}}`
-  const url = buildEndpoint(baseURL, {
-    tilesets,
-    query: encodeURIComponent(searchTerm),
-    limit: SEARCH_MODAL_PAGE_SIZE,
-    offset: page * SEARCH_MODAL_PAGE_SIZE,
+  // NOW - error 500
+  // https://api-dot-skytruth-pleuston.appspot.com/v2/tilesets/gfw-tasks-657-uvi-v2,gfw-tasks-658-reefer-uvi-v4/search/?query=orca&limit=14&offset=0
```
j8seangel: 🤕

src/app/search/searchActions.js
```diff
@@ -61,43 +59,88 @@ const loadSearchResults = debounce((searchTerm, page, state, dispatch) => {
...
+    })
+  })
+
+  const searchPromises = uniqSearchURLs.map((url) =>
+    fetch(url, options)
+      .then((response) => response.json())
+      .catch((err) => err)
+  )
+
+  Promise.all(searchPromises).then((resultsOrErrors) => {
```
j8seangel: I don't think this will return an array of all searchs with an error, following this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all#Rejection because if any of the promises fail it won't go go the `then` method but `catch` instead. Have you checked ?

src/app/vesselInfo/containers/VesselInfoPanel.js
```diff
@@ -9,32 +10,55 @@ import { setEncountersInfo } from 'app/encounters/encountersActions'
...
 
-  let layerFieldsHeaders
-  let layerIsPinable = true
-  if (currentlyShownLayer !== undefined && currentlyShownLayer.header !== undefined) {
-    if (currentlyShownLayer.header.info.fields !== undefined) {
-      layerFieldsHeaders = currentlyShownLayer.header.info.fields
-    }
-    if (currentlyShownLayer.header.pinable === false) {
-      layerIsPinable = false
+const getCurrentlyShownLayer = createSelector(
```
j8seangel: 👌 nice to use selectors where there haven't arrived yet.

src/app/vesselInfo/vesselInfoActions.js
```diff
@@ -206,13 +215,13 @@ export function hideVesselsInfoPanel() {
   }
 }
 
-export function addVessel({ tilesetId, seriesgroup, fromSearch = false, parentEncounter = null }) {
```
j8seangel: 👏 i like not couple the addVessel method and have another one to do something different

src/app/map/containers/MapWrapper.js
```diff
@@ -163,8 +185,9 @@ const mapStateToProps = (state) => ({
   loadTemporalExtent: state.filters.timelineOuterExtent,
   highlightTemporalExtent: state.filters.timelineOverExtent,
   // Internal
-  workspaceLayers: state.layers.workspaceLayers,
+  layerTitles: getLayersTitles(state),
   report: state.report,
+  isCluster,
```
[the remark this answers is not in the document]
nerik: I might be wrong, but `isCluster` is actually not anonymous, it is declared in the scope as a const. So we'll always have `isCluster === isCluster`

src/app/search/searchActions.js
nerik: Yeah this is meant to be removed - the whole multiple endpoints thing won't probably be necessary anyways.

src/app/search/searchActions.js
nerik: Yes. When one of the promises fail, it will go to the Promise.all `then`. Not sure why.

PR #1039, opened 2019-06-28
Adds a label to layer inputs so bablic can translate their names

satellitestudiodesign:
Adds a label with the same content as the input so bablic translates it, we show the label and hide the input. When user clicks "edit layers", we change the input value with the translated content of the label so user can edit from the right language.


j8seangel:
Great hack!

image

Tested and saved the workspace into Spanish, but then when you go back to the previous language only fishing efforts swap:
image


any idea why?

j8seangel:
It is not a bug as we want to preserve the translation for the workspace, but I have added a fallback to the loading event and seems it isn't always dispatched so doesn't always shows the language selector.

PR #1040, opened 2019-07-02
Rulers

nerik:
Adds the possibility to draw rulers on the map to check distances.
Needs https://github.com/GlobalFishingWatch/map-components/pull/65

Screenshot 2019-07-02 at 13 45 32

TODO:
- [x] check interaction with @satellitestudiodesign 
~- [ ] workspace save/read~
- [x] toolbar UI


src/app/rulers/components/Rulers.jsx
```diff
@@ -0,0 +1,43 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+import Icon from 'app/components/Shared/Icon'
+
+import RulersStyles from './rulers.module.scss'
+
+class Rulers extends Component {
+  state = {
```
j8seangel: 🙄 [the copy and paste components caveats]

src/app/rulers/rulersActions.js
```diff
@@ -0,0 +1,24 @@
+export const EDIT_RULER = 'EDIT_RULER'
+export const MOVE_CURRENT_RULER = 'MOVE_CURRENT_RULER'
+export const TOGGLE = 'TOGGLE'
```
j8seangel: Isn't this action too generic ? also we could have namespacing issues, so I'd prefer having something like `RULER_TOGGLE`

j8seangel: same for `RESET`

src/app/rulers/rulersSelectors.js
```diff
@@ -0,0 +1,50 @@
...
+  finalFeature.properties.isNew = isNew
+  return finalFeature
+}
+
+export const getRulersLayer = createSelector(
+  [getRulers, getRulersVisibility],
+  (rulers, rulersVisibility) => {
+    const geojson = {
+      type: 'FeatureCollection',
+      features: rulers.map(makeRulerGeometry),
```
j8seangel: Thanks for moving this logic to the selector, IMO it keeps it simpler and makes it easier to maintain

src/app/rulers/rulersActions.js
nerik: 👍

PR #1044, opened 2019-08-22
New timebar integration

nerik:
This replaces the legacy map-client Timebar with the `map-component` Timebar.

There are still some fixes to do (see https://github.com/GlobalFishingWatch/map-components/pull/81) but I think we should try to merge and test on develop.

TODO for later (check priority with Enrique):
- store bookmark in workspace
- store playback speed in workspace

Track view and speed/bearing/etc view will be done in another PR


src/styles/global.module.scss
```diff
@@ -25,7 +25,7 @@
 body,
 html {
   font-family: $font-family-1;
-  font-size: $font-size-normal;
+  font-size: 10px;
```
j8seangel: Doesn't this change many of things ? why is it needed ?

nerik: Because the timebar has `rem`s everywhere, so it will use the `font-size` defined in the body as "multiplier". In map-client, it was stupidly defined to something different than the browser defaults, so that's the only way I found to fix this, but am I overlooking something? (it has no impact because it seems everything in map client has an overridden font size anyways)

PR #1045, opened 2019-09-01
Timebar tracks and track features graphs

nerik:
Matching map-components PR: 
https://github.com/GlobalFishingWatch/map-components/pull/82

Screenshot 2019-09-04 at 17 11 58

Screenshot 2019-09-04 at 17 11 13

Notes:
- Only works with layers that have GeoJSON tracks available (Chile layers)
- Workspace layers need to have the following setting:
`"trackFeatures": ['course', 'speed']`
This is temporary as this should be set on headers (cc @andres-arana )
- API responses can be a bit slow (~10 seconds), see spinner on top left of the map for status


j8seangel:
LGTM

PR #1048, opened 2019-09-30
fixed filter style issues

nerik:
Fixes https://github.com/GlobalFishingWatch/GFW-Tasks/issues/1142
Fixes issue with out of focus filter values selector in filter modal.

### DIFF
--- a/src/app/filters/components/FilterGroupItem.jsx
+++ b/src/app/filters/components/FilterGroupItem.jsx
@@ -19,8 +19,6 @@ class FilterGroupItem extends Component {
   }
 
   onChangeFilterGroupVisibility() {
-    // TODO: hide quick edits and such
-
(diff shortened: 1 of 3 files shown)


src/app/filters/components/FilterGroupItem.jsx
```diff
@@ -63,35 +61,27 @@ class FilterGroupItem extends Component {
           hue={this.props.filterGroup.hue}
           onToggled={() => this.onChangeFilterGroupVisibility()}
         />
-        <input
+        <div
```
j8seangel: We can't use a div here, an input is needed for bablic to translate the content.

j8seangel: Sorry, this is my fault because the bablic translations only needs an input for custom layers and not by filters. Merging

PR #1049, opened 2019-10-21
config cloudbuild and run in cloud run

rrequero:
I've configured the literals files in the project (literals folder) with 2 files, one for each environment (dev and prod). The file is selected in the configuration of cloudbuild.


package.json
```diff
@@ -7,8 +7,9 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "serve": "http-server ./build",
-    "test": "react-scripts test",
+    "test": "npm run lint",
```
nerik: When is `test` ran? I'm not sure I'm comfortable having eslint running with `--fix` automatically.

rrequero: @nerik  In carrier portal and data portal we run this task before the build image task in cloudbuild. Same in this project. Maybe the fix should be executed in a pre-commit instead of the test stage.

nerik: >  Maybe the npm run lint with fix flag should be executed in a pre-commit instead of the test stage.

Personally I would prefer that, I think it's better to have human eyes reviewing linter changes (also an opportunity to learn)

rrequero: > > Maybe the npm run lint with fix flag should be executed in a pre-commit instead of the test stage.
> 
> Personally I would prefer that, I think it's better to have human eyes reviewing linter changes (also an opportunity to learn)

👍

nerik: Thanks 👍

rrequero: Done @nerik

literals/prod.json
```diff
@@ -0,0 +1,10 @@
+{
+  "fishing_hours_description": "Total fishing hours for all vessels on the selected time period are determined by the Global Fishing Watch fishing detection algorithm. Our fishing algorithm identifies each AIS broadcast as 'likely fishing' or 'not fishing.' We then assign an amount of time to each AIS point calculated from the time between each AIS broadcast. Because data from AIS is not evenly spaced in time,  the amount of time assigned to each AIS point varies.",
```
enriquetuya: @j8seangel and @nerik I've created a ticket to merge the to literals.json files the one from dev and prod. The only difference is the domain and we already have that in `REACT_APP_SITE_URL`. https://github.com/GlobalFishingWatch/GFW-Tasks/issues/1154

Dockerfile
```diff
@@ -0,0 +1,72 @@
+#################################################################################
+# Step to build the application assets
+#################################################################################
+FROM node:12 AS build
+RUN mkdir -p /app
+WORKDIR /app
+
+# Config build args
+ARG PUBLIC_URL=/
```
enriquetuya: @rrequero why do we have all these definitions in here? some of them are ok, but the ones that are environment dependant like `REACT_APP_MAP_URL`, `REACT_APP_V2_API_ENDPOINT`, `WORKSPACE_API_URL`, `REACT_APP_GOOGLE_TAG_MANAGER_KEY`, `REACT_APP_SHARE_BASE_URL`, `DEFAULT_WORKSPACE`, should be in cloud build level? Actually I would move every variable there, just to have consistency.

rrequero: @enriquetuya these are the args that cloudbuild replaces with the environment variables of the cloudbuild task when it generates the images. I configured by default these values.

Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+ARG REACT_APP_REQUIRE_MAP_LOGIN=false
+ARG REACT_APP_SHOW_BANNER=false
+ARG REACT_APP_WELCOME_MODAL_COOKIE_KEY=welcome_pop_up_v4
+ARG REACT_APP_TIMEBAR_DATA_URL=https://storage.googleapis.com/world-fishing-827/pelagos/data/vizzuality-map-resources/timebar
+ARG REACT_APP_V2_API_ENDPOINT=https://api-dot-skytruth-pelagos-production.appspot.com/v2
+ARG REACT_APP_BLOG_URL=http://blog.globalfishingwatch.org
+ARG REACT_APP_SITE_URL=http://globalfishingwatch.org/
+ARG REACT_APP_MAP_URL=http://globalfishingwatch.org/map
+ARG REACT_APP_SHARE_BASE_URL=https://globalfishingwatch.org/map/workspace/{workspace_id}
+ARG REACT_APP_GOOGLE_TAG_MANAGER_KEY=GTM-WXTMN69
```
enriquetuya: Remove this default var.

rrequero: done

PR #1051, opened 2019-10-29
open popup in new tab

nerik:
Open popup links in new tab as per ProtectedSeas team suggestion.


j8seangel:
Did you know about this https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/ ? I think it is a good practice when using target="_blank"

nerik:
@j8seangel Actually the linter flagged it but somehow I overlooked it. Thanks.

PR #1081, opened 2020-08-05
Encounters layers as a Mapbox GL layer

nerik:
Mapbox GL layer features which have `vessel_1_id` & `vessel_2_id` properties will trigger showing the encounters panel and load vessel info/tracks as with the pixi encounter layer.


nerik:
Additionally the info endpoint *for the encounter* is also failing with a 404 (only visible when loading a saved workspace)

